PR #203, opened 2017-05-31
Tutorial update

shih1:
Updated Tutorials 3 and 4 with simple text edits.

If we want to invest more time into tutorials then I can create a few more figures that may help.  I would focus on drawing diagrams so people understand


shih1:
Test

dkarchmer:
@shih1 don't forget to merge these changes

PR #204, opened 2017-06-02
automate rpc argument packing

diametralpitch:
TileBusProxyObject's rpc method can now support an optional kwarg 'arg_format'. If supplied, calling code will not have to manually pack a struct to be used as the argument to rpc, and instead, just pass the argument(s) normally and have the rpc method perform the struct packing based on the argument format supplied. Since the kwarg is optional, a conditional allows the code to fall back on the previous call to 'stream.send_rpc' so it will not break any existing code where the struct-packed argument is already given in advance to the rpc method.


timburke:
Let's release this as well so you see how to do that.

Releasing is done by:

1. Updating RELEASE.md in the affected packages (in this case `iotilecore`) is the only one
2. Updating version.py in the affected packages
3. pushing those changes and merging to master

Once all that is done and merged into master, you can trigger a new release by pushing a tag named `iotilecore-X.Y.Z` where the version has to match (and is checked by the deploy script) what is in RELEASE.md and version.py

timburke:
You can make the version number bump in this PR and then once I approve again and you merge then you can directly release a new iotile-core to pypi by:

```
git tag iotilecore-3.14.4
git push --tags
```

timburke:
There will be a slack message in coretools-users when it's out

diametralpitch:
Updated RELEASE.md and version.py

PR #209, opened 2017-06-06
avoid connection key error

diametralpitch:
Dictionary access inside of `_get_connection` was made safer, along with a slight refactor of calling code.


diametralpitch:
adjustments made. let me know if the logging makes sense.

timburke:
Looks good to me.  Style note that the % operator on the logging statement is not required but it doesn't hurt anything:

https://docs.python.org/2/library/logging.html#logging.Logger.debug

PR #265, opened 2017-09-12
Feat fleets

lelithium:
Adds `iotile cloud get_fleet {fleet_id}` to get information about a fleet's devices
Adds `iotile cloud get_whitelist {device_id}` to get a list of devices this device is supposed to manage
Unit tested

Also minor changes / cleanup in test_check_time function

### DIFF
--- a/iotile_ext_cloud/iotile/cloud/cloud.py
+++ b/iotile_ext_cloud/iotile/cloud/cloud.py
@@ -15,6 +15,7 @@ from iotile.core.dev.registry import ComponentRegistry
 from iotile.core.dev.config import ConfigManager
 from iotile.core.exceptions import ArgumentError, ExternalError
 from iotile.core.utilities.typedargs import context, param, return_type, annotated, type_system
+from utilities import device_id_to_slug
 
 @context("IOTileCloud")
 class IOTileCloud(object):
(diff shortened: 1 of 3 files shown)


iotile_ext_cloud/iotile/cloud/cloud.py
```diff
@@ -91,6 +92,46 @@ def device_info(self, device_id):
...
+    @param("fleet_id", "integer", desc="Id of the fleet we want to retrieve")
+    @return_type("basic_dict")
+    def get_fleet(self, fleet_id):
+        """ Returns the devices in the given fleet"""
+        api = self.api
+        if (fleet_id > pow(16,12) or fleet_id < 0):
+            raise ArgumentError("Fleet id outside of bounds")
+        slug = "g--" + device_id_to_slug(fleet_id)[8:]
+        try :
+            return api.fleet(slug).devices.get()
```
timburke: I would have this return a dictionary mapping device slugs to devices to make this function easier to work with.  This will simplify the `get_whitelist` function

```
results = api.fleet(slug).devices.get()
entries = results.get('results', [])

return {entry['device']: entry for entry in entries}
```

iotile_ext_cloud/iotile/cloud/cloud.py
```diff
@@ -91,6 +92,46 @@ def device_info(self, device_id):
...
+        try:
+            fleets = api.fleet.get(device=slug)['results']
+        except HttpNotFoundError:
+            raise ExternalError("Could not find the right URL. Are fleets enabled ?")
+
+        if not fleets:
+            # This is to be expected for devices set to take data from all project, or any device.
+            raise ExternalError("The device isn't in any network !")
+
+        out = {}
```
timburke: You can simplify the following logic to make it more understandable:

```python
networks = [self.get_fleet(fleet['id']) for fleet in fleets if fleet.get('is_network', False) is True]
networks_to_manage = [x for x in networks is x.get(slug, {}).get('is_access_point', False) is True]

out = {}
for network in networks_to_manage :
    out.update(network)

# Remove ourselves from the whitelist that we are supposed to manage
if slug in out:
   del out['slug']

return out

lelithium:
@timburke Only change I did from your req. changes was to remove the device_slug from the object, as it is already the key, and that seemed redundant

PR #335, opened 2018-02-05
Optimize _process_hex in DebugManager

x24ling:
Fixes issue #332


x24ling:
Huh, that's nice they had that. Updated IntelHex and used built-in function

PR #336, opened 2018-02-06
iotile-ship first PR

x24ling:


iotileship/iotile/ship/recipe.py
```diff
@@ -0,0 +1,76 @@
...
+        raise UnknownRecipeActionType("Unknown stage type specified, cannot create it", name=step_type)
+
+
+    @classmethod
+    def FromFile(cls, path, actions_dict, file_format="yaml"):
+        cls._actions_dict = actions_dict
+
+        format_map = {
+            "json": cls._process_json,
+            "yaml": cls._process_yaml,
```
dkarchmer: Do we really want/need to support two formats?

dkarchmer: Not a big deal from the coding stand point, but it creates room for one person doing one thing and another doing another, and for something like this, I am not sure there is a reason to allow for choice

x24ling: Good point, the preference is to support only yaml.

iotileship/iotile/ship/recipe_action.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import (unicode_literals, print_function, absolute_import)
+from builtins import str
+
+from iotile.core.exceptions import *
+
+class RecipeActionObject (object):
```
timburke: I think this entire file can be removed.  As discussed, recipe actions won't actually inherit from this class since they will be  defined in various places so I don't think this class will ever actually be used.

x24ling: yes, i agree

Issue #337, opened 2018-02-08
Add set_device_template command to iotile_ext_cloud


timburke:
There needs to be a function:

```
def set_device_template(self, device_id, template_name, os_tag=None):
    """Sets the device template for the given device in iotile.cloud.
        
        Is os_tag is passed, verify that the device template explicitly matches
        the expected os_tag by making another API call.
    """
```

added to `IOTileCloud` in `iotile-ext-cloud`

@dkarchmer What API is this a patch on?

timburke:
It should go in this file:
https://github.com/iotile/coretools/blob/master/iotile_ext_cloud/iotile/cloud/cloud.py#L85

dkarchmer:
I wonder if the API should also confirm that the os_tag matches that template_name. If so, you will want to check from:

https://iotile.cloud/api/v1/dt/pod-1gg1-v3-0-1/

But the API to patch is the same one as the sensor graph

PATCH: /api/v1/device
```
payload = {
    'template': 'slug'
}
```

timburke:
I would say we should add an optional argument that takes an `os_tag` and if given explicitly verifies it before patching and raises if it doesn't match.

dkarchmer:
May want to do the same on the old sensor_graph as well, and pass an optional app_tag

timburke:
Yup, agreed.  So to confirm there are now two parts of this:

- [ ] Add a new `set_device_template` function that take a device_id, a template name and optional `os_tag`
- [ ] Modify `set_sensorgraph` to take an optional `app_tag` and explicitly verify that if passed.

x24ling:
What's the best way to test these functions? It seems like currently there are tests to use a mock cloud.

The way i'm currently acquiring tags asscociated with the values is this way, but i always get a return where 'Authentication credentials were not provided'.
sg_os_tag = requests.get('https://iotile.cloud/api/v1/dt/%s/' % (new_template)).json().get('os_tag', None)

I also wrote test_set_sensorgraph, but it doesn't seem to be passing with the current mock cloud, is there something i should be setting? I'll be creating a pr to show you what i mean.

x24ling:
nvmd, i just found a good reference in test_check_time()

timburke:
The `mock_cloud` may need to be modified to allow for patching the relevant APIs.  I don't believe it has support for that right now.  The `mock_cloud` implementation is here:

[...]

PR #338, opened 2018-02-08
set_device_template. check ap/os tags

x24ling:
For #337

### DIFF
--- a/iotile_ext_cloud/iotile/cloud/cloud.py
+++ b/iotile_ext_cloud/iotile/cloud/cloud.py
@@ -181,7 +181,7 @@ class IOTileCloud(object):
    
         if app_tag is not None:
             if sg.get('app_tag', None) != app_tag:
-                raise ArgumentError("Cloud device template os tag mismatch", value=new_sg, cloud_sg_app_tag=sg.get('app_tag', None), app_tag_set=app_tag)
+                raise ArgumentError("Cloud sensorgraph record does not match app tag", value=new_sg, cloud_sg_app_tag=sg.get('app_tag', None), app_tag_set=app_tag)
 
         slug = device_id_to_slug(device_id)
         patch = {'sg': new_sg}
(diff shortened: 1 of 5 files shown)


x24ling:
Note, i'm will be fixing up the error raising logic, that isn't finalized.

iotile_ext_cloud/iotile/cloud/cloud.py
```diff
@@ -176,6 +175,31 @@ def set_sensorgraph(self, device_id, new_sg):
             else:
                 raise ExternalError("Error calling method on iotile.cloud", exception=exc, response=exc.response.status_code)
 
+    @param("device_id", "integer", desc="ID of the device that we want information about")
+    @param("new_template", "string", desc="The new device template that we want to set")
+    @param("os_tag", "integer", desc="Optional arg to check if the sensorgraph on the cloud matches the os_tag")
+    def set_device_template(self, device_id, new_template, os_tag=None):
+        """Sets the device template for the given device in iotile.cloud."""
+        if os_tag is not None:
+            sg_os_tag = requests.get('https://iotile.cloud/api/v1/dt/moo%s/' % (new_template)).json().get('os_tag', None)
```
dkarchmer: No reason to ever do a manual GET. This is also not checking if the template was found and the cloud had no issues returning the data

You should always use the python_iotile_cloud to do anything on the cloud:

```
try:
     dt = self.api.dt(new_template).get()
except RestHttpBaseException, exc:
     ...
```

under a try/catch just as you do when patching the device

iotile_ext_cloud/iotile/cloud/cloud.py
```diff
@@ -161,9 +161,8 @@ def check_time(self, max_slop=300):
 
     @param("device_id", "integer", desc="ID of the device that we want information about")
     @param("new_sg", "string", desc="The new sensor graph id that we want to load")
-    def set_sensorgraph(self, device_id, new_sg):
+    def set_sensorgraph(self, device_id, new_sg, app_tag=None):
```
dkarchmer: Where are you using this?

You need to do the same

```
try:
     sg = self.api.sg(new_sg).get()
except RestHttpBaseException, exc:
     ...
```

to get the sensor graph and then check that `sg['app_tag'] == app_tag`

x24ling:
To test this function, I also need to break out the mock api calls for sg and dt, I'll be adding that into the mock cloud as well.

PR #339, opened 2018-02-14
Add test case covering the use of service_delegate tile.

timburke:
The test case shows how to setup everything required to invoke an RPC on an RPCDisptacher subclass that is connected to an IOTileSupervisor.  The invocation is made using `HardwareManager` by creating an appropriate config file that sets up a virtual device pointing at the desired service.


timburke:
@diametralpitch Here's an example of how to properly set up a `HardwareManager` instance in a test fixture to point at a running service.

PR #349, opened 2018-02-20
Add support for matching apps to devices in HardwareManager

### LINKED ISSUE #303: Add support for app plugin to proxy objects in HWManager
There should be a way to match a whole device style plugin based on the contents of the app/os tags retrieved from the device's controller.

### LINKED ISSUE #348: Add support for waiting for a fixed amount of tracing data
Currently you can only dump tracing data, you can't synchronously wait for a fixed amount of data to come in.  This is important for tiles that use the tracing interface to send highspeed data out of an iotile device.

timburke:
### Overview

This PR adds the ability to load an `IOTileApp` object for a connected IOTile device in HardwareManager similar to how we are able to load proxy modules.  There is a new `app()` method on HardwareManager that attempts to find the best app matching a device based on its app_tag and version.

This will allow the creation of high level interfaces that apply to a whole device rather than just a single tile and will prevent needing to shoehorn cross-device functionality into a tile's proxy.

Also refactors and cleans up streaming and tracing code to allow `enable_tracing` and `enable_streaming` to be loaded multiple times per connection without breaking anything.

### Usage

To create your own app, you need to define a class that inherits from `IOTileApp` in `iotile.core.hw`.  See the docstring.  If your app matches best when someone calls `app` on a connected device, your class will be instantiated, passed the connected HardwareManager object and info and then returned to the user.

```python
class IOTileApp(object):
    """Base class for all IOTile Apps.

    An IOTile App is a class that represents all of the externally visible
    functionality of an IOTile device.  It is typically a subset of what is
    exposed on the individual tiles that make up the iotile device with a
    higher level API.  For example, there may be a shipping tracker IOTile
    device that has an accelerometer and an environmental tile.  The proxy
    objects for those tiles will expose a lot of functionality related to
    acceleration and environmental readings but nothing related to tracking
    a shipment.

    An IOTile App ties together the behavior of all of its underlying tiles.

    IOTile Apps are matched with devices using the device's app tag and version
    which is a number that unique identifies the app running on the device and
    a major.minor version number.

    All IOTile Apps are required to implement a MatchInfo module function if
    they support automatic matching with a device.  The MatchInfo function
    should return a list of (tag, SemanticVersionRange, quality) tuples for each
    app tag they should match with.  The quality value is a floating point
    number between 0 and 100, inclusive, where higher quality matches take
    priority over lower quality matches.  A good default choice is 50.

    Args:
        hw (HardwareManager): A HardwareManager instance connected to a
            matching device.
        app_info (tuple): The app_tag and version of the device we are
            connected to.
        os_info (tuple): The os_tag and version of the device we are
            connected to.
    """
```

Closes #303
Closes #348


dkarchmer:
(Sorry but not enough experience with the code to review)
Can you give an example of how you load/start the app within iotile cmd? Not obvious from the diff (without experience with the code):
`iotile hw --port=bled112 myapp`? 
Can there be a list of apps for a device? or can they only have one associated to the 
I assume you still see all other commands. e.g. "hw controller" or is this hidden somehow once an app is loaded?

timburke:
Yeah the way you would use it is:

```
iotile hw connect UUID app
(DeviceSpecificApp)
```

Similarly to how you can get a proxy object by calling `get(address)` or `controller`, what happens behinds the scenes is:
- HardwareManager has a list of installed apps (each one matches a specific app tag and version range).
- It gets the app_tag and version from the device and looks it up in its list of installed apps to see if it matches
- If multiple apps match the same app tag and version (not typical but you could imagine it) there is a tiebreaker with a 'quality' number so that the highest quality app class wins
- Every app also has a name in is `AppName()` classmethod.  If you know what app you want to load you can name it directly by using `app --name <name>`.  This allows having generic apps that can apply to many kinds of devices like a report uploader app or a device information app.  Those would not match automatically against any device but could be loaded by name.

timburke:
@dkarchmer @x24ling I added the first example of an app to `iotile-ext-cloud`.  It's called `cloud_uploader` and has a single method `upload` that will upload all data on a device to the cloud.

Usage is:

```
iotile hw connect <UUID> app --name cloud_uploader upload
```

If you want to watch its progress, turn on logging:

```
iotile -vvv hw connect <UUID> app --name cloud_uploader upload
```

The code is here fore reference:
https://github.com/iotile/coretools/blob/feat-load-apps/iotile_ext_cloud/iotile/cloud/apps/cloud_uploader.py

This is an example of making a simple app plugin to provide functionality across different iotile devices.

PR #351, opened 2018-02-20
Feat ship scripts

x24ling:
-iotile-ship script created
-replacement of variables set

#357, #358


iotileship/iotile/ship/scripts/iotile_ship.py
```diff
@@ -0,0 +1,93 @@
+"""Command line script to load and run a sensor graph."""
```
timburke: For python 3 compatibility all files should start with:

```
from __future__ import (unicode_literals, absolute_import, print_function)
from builtins import str
```

This will make sure all of the easy portability mistakes get caught and you don't need to prefix all of your strings with `u`

iotileship/iotile/ship/recipe.py
```diff
@@ -88,7 +89,33 @@ def _process_yaml(cls, yamlfile):
...
 
-    def prepare(self, variables=None):
+    def _complete_parameters(self, param, variables):
+        """Replace any parameters passed as {} in the yaml file with the 
+        variable names that are passed in
+        
+        Only strings, lists of strings, and dictionaries of strings can have 
+        replaceable values at the moment
+        """
+        if type(param).__name__ == 'str':
```
timburke: These explicit type checks will break on python 3 and on python 2 randomly if the string you get happens to be unicode.

I would check for dict and list and otherwise assume its a string and template it.

iotileship/iotile/ship/actions/pipe_snippet_step.py
```diff
@@ -0,0 +1,56 @@
...
+        self._expect = args.get('expect', None)
+
+        if self._expect is not None:
+            if len(self._expect) != len(self._commands):
+                raise ArgumentError("Length of expect must match length of commands", \
+                    parameter_name='commands')
+
+    def run(self):
+        process = Popen(shlex.split(self._context), stdout=PIPE, stdin=PIPE, stderr=STDOUT)
+        out, err = process.communicate(input='\r\n'.join(self._commands))
```
timburke: Just `\n` is all that is required here, not '\r\n'.

Issue #325, opened 2018-01-26
Add streamer acknowledgeable virtual device


timburke:
For testing streamer acknowledgements, we need a virtual device that implements enough of the relevant RPCs to be able to send reports and receive their acknowledgements.

timburke:
The easiest way to do this would be to modify `report_test_device.py` with support for 2 more RPCs.
https://github.com/iotile/coretools/blob/master/iotiletest/iotile/mock/devices/report_test_device.py

In particular the following changes would need to be made:

- [ ] An internal variable that tracks what acknowledgement value has been received
- [ ] in `open_streaming_interface`, if at least one acknowledgement has been received, starting the reading ids at `1 + last_acknowledgement_received`.  Otherwise always start them at `self.start_id`.
- [ ] Add a new rpc `acknowledge_streamer` as specified below
- [ ] Add a new rpc `query_streamer` as specified below
- [ ] Add a new argument that takes in a 6-character name and uses that as the module name (if passed) instead of `Simple`.  This is to allow for this device to impersonate the module name returned by a physical device so a proper proxy can be found for it.

The behavior should be that the test device will always produce `self.report_length` readings with ids that are sequential starting either at `self.start_id` or if an acknowledgement has been received, starting just after that acknowledged value.

## acknowledge_streamer rpc
- address: 8
- rpc_id: 0x200f
- parameter format: "HHL"
- param 1: index, the index of the streamer that is being acknowledged
- param 2: force, 0 if this acknowledgement should only be applied if greater than what was already received, 1 to force rolling back the device's internal acknowledgement.
- param 3: acknowledgement, the last reading id known to be successfully received by the cloud
- return format: "L"
- return value: An error code, where 0 mean no error

Internally the device should store a map of `index, ack` and accept any index but it should only use the value at index 0 when computing what reading ids to send

## query_streamer rpc
- address: 8
- rpc_id: 0x200a
- parameter format: H
- param 1: index, the index of the straemer being queries
- return format: LLLLBBBx
- return value, a packed binary structure containing information on the streamer being queried by its index.  In this implementation a constant template be returned:
[0, 0, 0, self.acks.get(index, 0), 0, 0, 0]
where `self.acks` refers to the dictionary updated by calls to `acknowledge_streamer`.  The rest of the parameters are internal details of the streamer state machine that are not necessary to reproduce faithfully.

diametralpitch:
adressed in https://github.com/iotile/coretools/pull/353

PR #353, opened 2018-02-20
Feat streamer ack virtual device

diametralpitch:
addresses https://github.com/iotile/coretools/issues/325


iotiletest/iotile/mock/devices/report_test_device.py
```diff
@@ -43,9 +43,11 @@ class ReportTestDevice(VirtualIOTileDevice):
                 stream_id (int): The stream that this reading should be sent out
     """
 
-    def __init__(self, args):
+    def __init__(self, args, module_name=None):
```
timburke: I would put `module_name` as a parameter in args (with the default of 'Rptdev' if not passed) since I don't think there's any way to pass arguments to this constructor except via the args dictionary.

diametralpitch:
closes #325

Issue #350, opened 2018-02-20
Add support for creating bootstrap images in iotile-build


timburke:
There should be a way to combine multiple elfs into a single bootstrap `.hex` file.

The functionality should live in:
https://github.com/iotile/coretools/blob/master/iotilebuild/iotile/build/config/site_scons/autobuild.py

There should be a new function like:

```python
def autobuild_bootstrap_file(file_name, image_list):
    """Combine multiple firmware images into a single bootstrap hex file."""
```

All of the files in image_list should be assumed to live in `build/output` and `file_name` should be generated in that folder as well.  You can look at `autobuild_release` for examples of how to set up the scons environment for building:

https://github.com/iotile/coretools/blob/master/iotilebuild/iotile/build/config/site_scons/autobuild.py#L104

x24ling:
Closed by PR #354

PR #354, opened 2018-02-20
Feat build bootstrap

x24ling:
For #350


iotilebuild/iotile/build/config/site_scons/autobuild.py
```diff
@@ -211,3 +211,40 @@ def autobuild_documentation(tile):
...
+        file_format = ext[1:]
+        file = os.path.join(outputbase, image)
+
+        if file_format == 'hex':
+            full_image_list_names.append(file)
+        elif file_format == 'elf':
+            new_file = file.replace('.elf','.hex')
+            outhex = env.Command(new_file, file, "arm-none-eabi-objcopy -O ihex $SOURCE $TARGET")
+            full_image_list_names.append(new_file)
+            Delete(new_file)
```
timburke: This line won't do what you expect it to because it isn't attached to a command that has a source and target so SCons doesn't know when to run it.  What you should do is keep track of all of the `new_files` that you need to make here and append them to the list of actions in the final `env.Comand` that you're creating at the end of the function as in:
http://scons.org/doc/1.2.0/HTML/scons-user/x2940.html

You'll have to build the list of actions step by step and then pass the list as a whole to `env.Command` instead of passing `action=`.  That way these temporary files will automatically be deleted as part of the build process after merging them.  You'll need to modify the test code to not reference the .hex files as well.  This will prevent polluting `build/output` with temporary `.hex` files.

iotilebuild/test/test_iotilebuild/test_scons.py
```diff
@@ -108,3 +110,26 @@ def test_build_prerelease(tmpdir):
...
+    try:
+        os.chdir(builddir)
+        err = subprocess.check_call(["iotile", "build"])
+        assert err == 0
+
+        print os.listdir(os.path.join('build','output',builddir))
+
+        hexdata = IntelHex(os.path.join('build','output', 'test_hexfile.hex'))
+        assert hexdata.segments() == [(0x10001014, 0x10001018)]
+        hexdata = IntelHex(os.path.join('build','output', 'test_elffile.hex'))
```
timburke: You can remove this line once you refactor the build step to not leave temporary hex files.

iotilebuild/iotile/build/config/site_scons/autobuild.py
```diff
@@ -211,3 +211,40 @@ def autobuild_documentation(tile):
...
+        if len(ext) == 0:
+            raise ArgumentError("Unknown file format or missing file extension", file_name=image)
+        file_format = ext[1:]
+        file = os.path.join(outputbase, image)
+
+        if file_format == 'hex':
+            full_image_list_names.append(file)
+        elif file_format == 'elf':
+            new_file = file.replace('.elf','.hex')
+            outhex = env.Command(new_file, file, "arm-none-eabi-objcopy -O ihex $SOURCE $TARGET")
```
timburke: `outhex` is unused so you don't need to assign it to a variable

Issue #324, opened 2018-01-26
Add function to get all streamers for a given device


timburke:
Currently we only have `highest_acknowledgement` which takes a streamer and index and gets the acknowledgement value for that streamer.  We need one that gets all acknowledgements for a given device.

The signature should be:

```python
def device_acknowledgements(self, device_id):
"""Get all streamer acknowledgements for a device by its id.

Args:
    device_id (int): The device we are querying

Returns:
    list of namedtuples: A list of all acknowledgement values received from the cloud.  
        The namedtuples should have index, ack and selector fields pulled from the corresponding
        record in the cloud.
```

The API to use is `/api/v1/streamer/?device=<slug>`.  A single call is all that is needed.  The `ack` value is stored in the `last_id` of the cloud response.  The index should be the value stored as `index` in the individual result, not the index of that result in the list of results.  They will typically be the same but don't have to be in general.

For example:
https://iotile.cloud/api/v1/streamer/?device=d--0000-0000-0000-0036
```json
{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 242,
            "slug": "t--0000-0000-0000-0036--0000",
            "device": "d--0000-0000-0000-0036",
            "index": 0,
            "last_id": 27282,
            "last_reboot_ts": "2017-08-08T19:30:37Z",
            "is_system": false,
            "selector": 22527
        },
        {
            "id": 243,
            "slug": "t--0000-0000-0000-0036--0001",
            "device": "d--0000-0000-0000-0036",
            "index": 1,
            "last_id": 27283,
            "last_reboot_ts": "2017-08-08T19:30:37Z",
            "is_system": true,
            "selector": 24575
        }
    ]
}
```

diametralpitch:
done in https://github.com/iotile/coretools/pull/355

PR #355, opened 2018-02-21
Feat device acknowledgements

diametralpitch:
addresses https://github.com/iotile/coretools/issues/324


diametralpitch:
closes #324

PR #384, opened 2018-03-04
Upgrade to using iotile_cloud v0.8.11

dkarchmer:
Which requires IOTileDeviceSlug users to pass a `allow_64bits=False` if
they want a check for 48bit limit. If this change is not done, test
will fail next time somebody pip installs v0.8.11. No need to release
this change


dkarchmer:
Ugly change, but my previous change to iotile_cloud has some issues with DataBlocks, so IOTileDeviceSlug needs to accept 64bits by default, even though a device itself can never use more than 48bits

PR #403, opened 2018-03-09
Add device_updater script

### LINKED ISSUE #390: Add IOTileApp that can run a script on an iotile device
Now that we have the ability to generate and parse `UpdateScript` objects in CoreTools, we also need an easy way to run those scripts on an iotile device.

There should be an `IOTileApp` named `script_updater` that has the ability to download and run a script on an IOTile device.

timburke:
## Overview

This PR adds a `device_updater` IOTileApp.  This app just has a single user accessible function from the `iotile` tool:

```python
@docannotate
    def load_script(self, script_path, confirm=True):
        """Load a script from a file and run it.

        This function will load a binary update script from the file given in
        script_path, download it to the device, run it and wait for it to
        finish.

        Any errors during script execution will be raised as HardwareError
        exceptions.

        Args:
            script_path (str): The path to the binary script file (.trub extension)
                that you wish to load.
            confirm (bool): Require explicit confirmation from an interactive user
                before running the script on the device.  This defaults to true,
                which means if you want to use this function in a batch setting.
                You need to pass confirm=False.
        """
```

If you have an updating script, this app will let you program it.  Usage is:

```
> iotile hw connect <device> app --name device_updater load_script <path to script>

```

This will synchronously read the script and prompt the user with a few questions to make sure you really want to update your device.  You can optionally see the script that you will be programming.  Example usage:

```
Loaded script with 1 actions from file .\tile.trub
Running script on device 0x532 with app info (2049 1.0.0) and os_info (0 0.0.0)
Do you want to see the script contents before programing (y/n)? y

Script Actions:
01: Reflash slot 2 with 26624 (0x6800) bytes starting at offset 6144 (0x1800) (hardware_type: 0)

Are you sure you want to program this script (y/n)? y
Downloading script: [########################################] 6.1 seconds total
Waiting for script to validate
Waiting for script to finish executing
...................................................................................
```

The function will wait for the script to finish execution before returning even if that requires waiting for the device to reset itself.

## Usage From Python
If you want to use this function from a python script, you probably don't want the user facing `load_script` function so there is an internal `run_script` function that is designed to be called programmatically.

```python
def run_script(self, script, force=False):
        """Run a script on the connected IOTile device.

        The script must an UpdateScript object.  If you are looking for an
        interactive function that will load a script from a file and show
        update information to a user, you should use load_script. This is a
        low level function that is meant to be called from a larger program.

        If this run is called from an interactive session (inside the iotile
        tool) then it will show a progress bar during download and periodic
        updates during script execution, otherwise there is no feedback until
        the script is finished.

        Args:
            script (UpdateScript): The script that we wish to run on our attached
                device.
            force (bool): If there is already a script loaded but not yet executed,
                clear that script before proceeding.
        """
```

Closes #390


timburke:
I agree with wanting to be able to download the script automatically.  I added that feature to #404 since that app goes in iotile-ext-cloud so it has access to cloud APIs.

PR #406, opened 2018-03-14
WebSocketDeviceAdapter - First PR for approval

haroal:
There are a lot of files in the PR because while I was reading the code to understand how it works, I fixed some spelling mistakes or not Python2/3 compatible features (in Exception for instance). My real work is in `transport_plugin/websocket` (I don't know if it is there that I have to put this, but it seems me correct as there already is awsiot, and I can use my `WebSocketDeviceAdapter` through the `iotile` command correctly :) )

I didn't do the `WebSocketVirtualInterface` as you told me not to do, so it was a bit harder to test, especially because the new `ValidatingWebSocket` client is not compatible with the old `WebSocketHandler`... For now, I did what I was able to do without writing the `WebSocketHandler2`, to make a "smaller" PR. 

I'm just waiting for your approval to know if I am on a good way before continuing and starting the `WebSocketHandler2`. Then, I will go back to the `WebSocketDeviceAdapter` to make it fully work, with good async methods...


transport_plugins/websocket/iotile_transport_websocket/device_adapter.py
```diff
@@ -0,0 +1,85 @@
...
+                self.disconnect_sync(connection)
+            except HardwareError:
+                pass
+
+        self.client.stop()
+        self.connections.stop()
+
+    def periodic_callback(self):
+        try:
+            self.client.send_ping()
```
timburke: I appreciate the intent here but I would remove this for now and if you want to add it back in, it will need to be nonblocking since periodic callback for device adapters in the `iotile-gateway` will be called in the main event loop as a periodic callback so they can't block the entire loop with a possible 10 second timeout. 

I wonder if this is necessary since the websocket client should flag us on disconnection anyway since it knows when the tcp socket goes away.  If you feel it is necessary then it would need to be done in a nonblocking way where, for example, you check when the last time you heard from the remote server was and if it's been longer than a fixed time you send a ping and set a flag that you have pinged and then the next periodic_callback if you have sent a ping and still haven't heard back you disconnect and stop.

haroal: Indeed, the `closed` function in `ValidatingWebSocket` is called if the server unexpectedly disconnect but there is nothing triggered to display something to the user as this `closed` function is usually used to wait for a normal disconnection...

I remove my code in `periodic_callback` because indeed it is not async(I forgot to add a FIXME there ^^), but I think we should handle the case where we are unexpectedly disconnected from the gateway, to make it close properly the connections and to warn the user

Issue #405, opened 2018-03-12
Promote websocket stream to DeviceAdapter


timburke:
## Overview

Right now, we have the ability to connect to an iotile device over websockets using:
https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/hw/transport/websocketstream.py

This is the client side of the connection, expecting a websocket server that is implemented here:
https://github.com/iotile/coretools/blob/master/iotilegateway/iotilegateway/wshandler.py

This is a very simplistic implementation that does not support multiple device connections at a time.  There is an implicit global `connection` in `wshandler.py` for the single connected device and messages from the client to the server do not specify what connection they pertain to:
https://github.com/iotile/coretools/blob/master/iotilegateway/iotilegateway/wshandler.py#L10

We want to promote `WebSocketStream` to `WebSocketAdapter` which is a full DeviceAdapter like `BLED112Adapter` and `AWSIOTDeviceAdapter`.  

## Background

The main differences between a `DeviceAdapter` and a `CMDStream` are:

- A `DeviceAdapter` has asynchronous routines that take a callback, whereas CMDStream routines are all synchronous.  This is because DeviceAdapters are designed to coexist well with multiple adapters in the same event loop process as in `iotile-gateway`.

- A `DeviceAdapter` can have multiple connections at once so every message to a connected device needs to include a `connection` key that tells the `DeviceAdapter` what specific connection it pertains to.  In `CMDStream`, there is a single global connection that all messages implicitly pertain to.

## Key Requirements

- This will require backwards incompatible changes to the messages passed between `WebSocketHandler` and `WebSocketAdapter` so `WebSocketGatewayAgent` should register two handlers: the current `WebSocketHandler` at path `/iotile/v1` and `WebSocketHandler2` at path `iotile/v2`.  This will allow the gateway to respond to messages from both kinds of clients without breaking current code.

- For testing and general usefulness, there should also be a `WebsocketVirtualInterface` that can serve a virtual IOTile device over websockets.  It can be modeled after the virtual `BLED112` interface:
https://github.com/iotile/coretools/blob/master/transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
All virtual interfaces are required to inherit from `VirtualInterface` and are used by the `virtual_device` script included with `iotile-core` to serve access to a virtual device:
https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/hw/virtual/virtualinterface.py

[...]

PR #411, opened 2018-03-20
WebSocketDeviceAdapter

haroal:
See issue #405 to understand the story of this PR.

I had a virtual interface which allows me to implement all the features needed by the DeviceAdapter (connecting, disconnecting, opening interfaces, streaming reports, tracing data, sending scripts, calling RPC, ...), in `WebSocketDeviceAdapter`. All these functions are now asynchronous.

I also had the `GatewayAgent` and the `WebSocketHandler` to add the server part into the gateway.

Finally, at the end, I mainly rewrite the `WebSocketVirtualInterface` to match more the `WebSocketHandler`. 

All the parts used the same 'protocol' to communicate. It is fully defined in a protocol folder.
  
---------
 
The only part which remains is to add unit tests to cover the code.


transport_plugins/websocket/iotile_transport_websocket/device_adapter.py
```diff
@@ -1,28 +1,47 @@
...
+        # Configuration
         self.set_config('default_timeout', 10.0)
         self.set_config('expiration_time', 60.0)
         self.set_config('probe_supported', True)
         self.set_config('probe_required', True)
+        self.mtu = int(self.get_config('mtu', 60*1024))  # Split script payloads larger than this
 
-        path = "ws://{0}/services".format(port)
+        # Set logger
+        self.logger = logging.getLogger('ws.manager')
```
timburke: Standard logging practice is to use the name of the module rather than a specific name so that it's easy for a consume to know how to turn loggers on and off by name:

```
self.logger = logging.getLogger(__name__)
```

The code in `bled112` transport plugin is old and contains this bad practice as well.  Here's a good reference on python logging best practices:
https://fangpenlin.com/posts/2012/08/26/good-logging-practice-in-python/

transport_plugins/websocket/iotile_transport_websocket/device_adapter.py
```diff
@@ -102,6 +155,281 @@ def stop_sync(self):
...
+        try:
+            self.connections.get_connection_id(connection_id)
+        except ArgumentError:
+            callback(connection_id, self.id, False, "Could not find connection information", None, None)
+            return
+
+        self.connections.begin_operation(connection_id, 'rpc', callback, timeout)
+
+        try:
+            result = self.client.send_command('send_rpc', {
```
timburke: As discussed yesterday, this is not truly async so you would need to post a message and then have the rest of this routine run in a callback on receiving the response back (timeout is handled by the connection manager).

transport_plugins/websocket/iotile_transport_websocket/device_adapter.py
```diff
@@ -102,6 +155,281 @@ def stop_sync(self):
...
+            callback(connection_id, self.id, False, "Could not find connection information")
+            return
+
+        context['progress_callback'] = progress_callback
+
+        self.connections.begin_operation(connection_id, 'script', callback, self.get_config('default_timeout'))
+
+        # Count number of chunks to send
+        nb_chunks = 1
+        if len(data) > self.mtu:
```
timburke: I would not cache `self.mtu` but rather request it from `get_config` once per function that needs it since otherwise there is no opportunity for anyone to override the default value because it gets cached into `self.mtu` in `__init__` and self.mtu is never updated again.

transport_plugins/websocket/iotile_transport_websocket/device_adapter.py
```diff
@@ -102,6 +155,281 @@ def stop_sync(self):
...
+            self.connections.get_connection_id(connection_id)
+        except ArgumentError:
+            callback(connection_id, self.id, False, "Could not find connection information")
+            return
+
+        self.connections.begin_operation(connection_id, 'open_interface', callback, self.get_config('default_timeout'))
+
+        try:
+            result = self.client.send_command('open_interface', {'interface': interface})
+            messages.OpenInterfaceResponse.verify(result)
```
timburke: Similar, this is not fully async as discussed.

transport_plugins/websocket/iotile_transport_websocket/device_adapter.py
```diff
@@ -78,31 +168,539 @@ def probe_async(self, callback):
...
 
-        result = self.client.send_command('scan', {})
-        if not result['success']:
-            callback(self.id, False, 'Error while scanning through ws : {}'.format(result['reason']))
-        else:
-            # FIXME: async after creating WebSocketHandler2
-            for dev in result['devices']:
-                self._trigger_callback('on_scan', self.id, dev, self.get_config('expiration_time'))
-            callback(self.id, True, None)
+        if self.probe_callback is not None:
```
timburke: There can be multiple async probes in flight at the same time from different clients so this is going to need to get enhanced at some point.  It's okay to leave for now but please file an issue so we don't forget about it.

transport_plugins/websocket/iotile_transport_websocket/virtual_websocket.py
```diff
@@ -0,0 +1,922 @@
...
+            chunk (bytes): A chunk that should be sent instead of requesting a
+                new chunk from the pending reports.
+        """
+
+        connection_string = self._uuid_to_connection_string(device_uuid)
+
+        self.streaming_data = True
+
+        if chunk is None:
+            chunk = self._next_streaming_chunk(20)
```
timburke: You can increase the transmit size here. The 20 bytes for the BLE virtual interface was because the notification packets used over BLE only support up to 20 bytes, here you could use your much larger MTU up to 64kb or whatever you like really.

transport_plugins/websocket/iotile_transport_websocket/virtual_websocket.py
```diff
@@ -0,0 +1,922 @@
...
+            chunk (bytes): A chunk that should be sent instead of requesting a
+                new chunk from the pending reports.
+        """
+
+        connection_string = self._uuid_to_connection_string(device_uuid)
+
+        self.tracing_data = True
+
+        if chunk is None:
+            chunk = self._next_tracing_chunk(20)
```
timburke: Same comment, this chunk size can go up dramatically.

transport_plugins/websocket/iotile_transport_websocket/virtual_websocket.py
```diff
@@ -0,0 +1,922 @@
...
+                operations.NOTIFY_TRACE,
+                connection_string=connection_string,
+                payload=base64.b64encode(chunk)
+            )
+            self._defer(self._send_trace, [device_uuid])
+        except HardwareError as err:
+            return_value = err.params['return_value']
+
+            # If we're told we ran out of memory, wait and try again
+            if return_value.get('code', 0) == 0x182:
```
timburke: This code can be removed.  The `0x182` error code is a specific firmware response from the BLED112 dongle that it doesn't have any more free internal 20 byte buffers to queue a notification for transmission on the next ble connection event so we have to retry. 

That can never happen with websockets so this code can probably be significantly simpler since you don't have to worry about retries.

haroal:
Closes #405

PR #419, opened 2018-04-02
replace subtract_a_from_b with subtract_afromb, like in lib_controller

x24ling:
#418

### DIFF
--- a/iotilesensorgraph/RELEASE.md
+++ b/iotilesensorgraph/RELEASE.md
@@ -2,6 +2,10 @@
 
 All major changes in each released version of iotile-sensorgraph are listed here.
 
+## 0.6.4
+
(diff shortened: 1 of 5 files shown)


x24ling:
Also, I got the statement to work by doing this
```
every 1 second
    {
        call 0x8010 on slot 1 => unbuffered 11;
        call 0x8020 on slot 1 => unbuffered 15;
        on unbuffered 15
        {
            subtract constant 13 => unbuffered 13, default 0;
        }
    }
```
otherwise, it would take the output of the node created by every 1 second

timburke:
That makes sense.  I added an issue for myself to clarify if that's a bug or intended behavior but that syntax will work regardless, I just thought there was a shorter way but it may not work yet

PR #434, opened 2018-04-09
Fix regression in ProductResolver

### LINKED ISSUE #433: Regression in iotile-build building projects with tilebus definitions or include_directories
The `ProductResolver` does not support `module_settings.json` files with arrays of values for a given key.

timburke:
Closes #433

### DIFF
--- a/iotilebuild/RELEASE.md
+++ b/iotilebuild/RELEASE.md
@@ -2,6 +2,11 @@
 
 All major changes in each released version of IOTileBuild are listed here.
 
+## 2.6.5
+
+- Fix regression building firmware with tilebus_definition or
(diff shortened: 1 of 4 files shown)


x24ling:
Is your current implementation just to skip adding the "include_directories" and "tilebus_definitions" to the ProductResolver?

timburke:
Yeah, there's not a use case right now for resolving those from other products so it was the quickest solution

timburke:
Yup, sorry that didn't get pushed.  Thanks for catching that.

timburke:
I added https://github.com/iotile/coretools/issues/435 to track a future enhancement where we support resolving tilebus definitions and include_directories

Issue #438, opened 2018-04-11
iotile-sensorgraph - incorrect node_sort behavior


x24ling:
For example:
```
every 10 minutes
{
    call 0x8000 on slot 4;
    call 0x8001 on slot 4 => output 35;
    call 0x8002 on slot 4 => output 33;
    call 0x8003 on slot 4 => output 34;
}
```
Before toposort:
```
0 : (system input 2 always) => counter 1024 using copy_all_a
1 : (system input 3 always) => counter 1025 using copy_all_a
2 : (system input 5 always) => counter 1026 using copy_all_a
3 : (system input 6 always) => counter 1027 using copy_all_a
4 : (counter 1024 when count >= 1) => counter 1028 using copy_latest_a
5 : (counter 1028 when count == 1 && constant 1024 always) => unbuffered 1024 using call_rpc
6 : (counter 1028 when count == 1 && constant 1025 always) => output 33 using call_rpc
7 : (counter 1028 when count == 1 && constant 1026 always) => output 34 using call_rpc
8 : (counter 1028 always) => counter 1029 using copy_all_a
9 : (counter 1029 when count == 1 && constant 1027 always) => output 35 using call_rpc
10 : (counter 1024 when count >= 360) => counter 1030 using copy_latest_a
11 : (system input 0 always && counter 1030 when count == 1) => system output 0 using copy_latest_a
```

And deptree
```
{0: set([]), 1: set([]), 2: set([]), 3: set([]), 4: set([0]), 5: set([4]), 6: set([4]), 7: set([4]), 8: set([4]), 9: set([8]), 10: set([0]), 11: set([10])}
```

nodeorder = 
```
[0, 1, 2, 3, 4, 10, 5, 6, 7, 8, 11, 9]
```

snippet created (i annotated the node indeces from before the toposort)
```
0: add_node "(system input 2 always) => counter 1024 using copy_all_a"
1: add_node "(system input 3 always) => counter 1025 using copy_all_a"
2: add_node "(system input 5 always) => counter 1026 using copy_all_a"
3: add_node "(system input 6 always) => counter 1027 using copy_all_a"
4: add_node "(counter 1024 when count >= 1) => counter 1028 using copy_latest_a"
10: add_node "(counter 1024 when count >= 360) => counter 1030 using copy_latest_a"
5: add_node "(counter 1028 when count == 1 && constant 1024 always) => unbuffered 1024 using call_rpc"
[...]

PR #444, opened 2018-04-18
Sg input buffered

x24ling:
For #438


x24ling:
https://app.iotile.cloud/#/shipping/d--0000-0000-0000-0531

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -355,6 +355,16 @@ def sort_nodes(self):
         node_order = toposort_flatten(node_deps)
         self.nodes = [self.nodes[x] for x in node_order]
 
+        #Check inputs all in the beginning
+        root_node_section_passed = False
+        for node in self.nodes:
+            if not root_node_section_passed:
+                if 'input' not in str(node):
```
timburke: There is a better way to do this check.  Checking the converted string is fragile because it relies on ancillary behavior that could change at any time.  For example it will break if we ever introduce a processing function that has the word input in its name.

The correct check is to iterate over the inputs of the node and check if `input[0].selector.input` is True.  You should do this check instead of anywhere where you are currently checking for the literal 'input' in a string

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -355,6 +355,16 @@ def sort_nodes(self):
...
+                    root_node_section_passed = True
+            else:
+                if 'input' in str(node):
+                    raise NodeConnectionError("Inputs not sorted in the beginning")
```
timburke: You should include the node and its position as keyword arguments in this error so its easy to identify which node caused the problem.

iotilesensorgraph/iotile/sg/parser/statements/copy_statement.py
```diff
@@ -95,7 +95,13 @@ def execute(self, sensor_graph, scope_stack):
             op = 'copy_count_a'
 
         if self.explicit_input:
-            sensor_graph.add_node(u"({} always && {} {}) => {} using {}".format(self.explicit_input, trigger_stream, trigger_cond, self.output, op))
+            #If root node is an input, create an intermediate node with an unbuffered node
+            if 'input' in str(self.explicit_input):
```
timburke: Again, this should check the type of the selector.

iotilesensorgraph/iotile/sg/parser/statements/on_block.py
```diff
@@ -121,7 +121,16 @@ def execute_before(self, sensor_graph, scope_stack):
             else:
                 combiner = '||'
 
-            sensor_graph.add_node(u"({} {} {} {} {}) => {} using copy_latest_a".format(stream_a, trigger_a, combiner, stream_b, trigger_b, trigger_stream))
+            if 'input' in str(stream_a) and 'input' not in str(stream_b):
```
timburke: Same, check for selector

iotilesensorgraph/iotile/sg/parser/statements/copy_statement.py
x24ling: What's the difference between a DataStream and DataStreamSelector? Sometimes, self.explicit_input is just a DataStream so it doesn't have a selector variable. They both do have a .stream_type variable which I can use to check if it's a DataStream.InputType.

iotilesensorgraph/iotile/sg/parser/statements/on_block.py
x24ling: stream_a and stream_b are both DataStream objects so they don't have a selector. I could if the stream_type variable is a DataStream.InputType. 

If this is the case, I will add property function 'input' into the DataStream object that's similar to output:
https://github.com/iotile/coretools/blob/master/iotilesensorgraph/iotile/sg/stream.py#L70

timburke: Correct, yeah you should add that property

iotilesensorgraph/iotile/sg/parser/statements/copy_statement.py
timburke: DataStream can only name a single stream like `input 15`.  A selector is a criteria used to select streams so it could select a single stream in which case it's the same `input 15' but it could select a whole class of streams like `all inputs`.

x24ling:
When you say "prepended", do you mean the list of continuous root nodes that would exclude the one root node that is not at the beginning of the the list of nodes? For example:

input 1 => output 1
unbuffered 1 => unbuffered 2
input 2 and unbuffered 2 => unbuffered 3

The prepended nodes would just be the first node? That way, process_input would only check the first node and ignore the third node?

Unless I'm misinterpreting what process_inputs does, I think the actual sensor_graph on the controller does something different. It still sees the third node and will process it, it's just what will happen is that the nodes on the controller tile will be switched as below:

input 1 => output 1
input 2 and unbuffered 2 => unbuffered 3
unbuffered 1 => unbuffered 2

What this breaks is that the second node will no longer be functional because it depends on unbuffered 2 in the third node.

If the goal is to ensure that improperly sorted sensorgraphs are simulated as such, I'm not sure if that's needed because of the check in sort_nodes() that prevents improperly sorted sensorgraphs from being made in the first place. Unless there's something else I'm missing?

timburke:
Yeah fair enough.  The intent was to make the simulator work the same as the device but you're right the correct way to simulate it is not what I said for the reason you gave.  Let's drop that idea for now

x24ling:
Out of curiosity, when we push this new version of iotile-sensorgraph and we upgrade the coretools_serial.txt in production_devices, do we also need to upgrade the app_tag version of every project that processes a .sgf?

PR #463, opened 2018-06-20
Feat multi jlink

### LINKED ISSUE #462: Enable multiplex jlink connection
This will require
For implementing the multiplexed jlink:

`iotile-core`:
1. Update hwmanager.py so that `debug` command takes an optional string parameter with a "connection_string" that can be interpreted as the multiplexer channel
2. Update streamadapter.py so that `debug` command passes this optional parameter
3. Update adapter.py so that `open_interface_async` takes an optional arg and passes it to `open_debug_interface`
4. Update adapter.py so that `open_debug_interface takes an optional param`

`iotile-transport-jlink`:
1. Update argument parsing for the `jlink` transport adapter to take a `mux=<some string>` to indicate that we are talking to a jlink with a multiplexer and allow us to change the multiplexer design in the future
1. Update `open_debug_interface`to use the argument to set multiplexer settings
2. Update `connect_async`to look for the same string as in `open_debug_interface`to decide what channel to use

`FlashStep`:
1. Add an additional param to indicate what multiplexer channel to use that is optional

x24ling:
Closes #462


x24ling:
Comment on bled112:
```debug``` currently doesn't work because bled112 doesn't currently have ```_open_debug_interface``` implemented, so the error below is thrown (which is correct). 


```
AttributeError 'BLED112Adapter' object has no attribute '_open_debug_interface'
```

Issue #473, opened 2018-06-28
jlink transport plugin debug fails when calling it a second time


x24ling:
Need to close jlink if connecting to a different mux channel.

PR #474, opened 2018-06-28
Call stop sync before going into debug interface

x24ling:
For issue #473


timburke:
That seems like a workable strategy

x24ling:
OK, here is one wierd case:

```
connect_direct device=nrf52;channel=0  #Establishes new connection with channel 0
debug  #Quickly connects to debug properly
debug -c device=lpc824;channel=1 #Establishes new connection with channel 1
back
controller #Fails. Still thinks it's connected although the jlink is connected to channel 1.
```

I think it might be good to block changing debug connnection ports if a connect_direct line has been set, thoughts?

x24ling:
Does it need to be a dictionary? I just created a variable self._connection_id that get's stored in the JLinkAdapter object, and I believe it works

timburke:
Yeah that's fine.  In general it needs to be a dictionary for device adapters that support multiple connections at he same time but this one doesn't so a single variable is fine.

x24ling:
OK, I got the programming time from 44 seconds to 24 seconds by using shared resources. I think there's a little bit more optimization that can be done, but it's looking much better. 

I will need update three packages, iotile-core, iotile-ship, and transport-jlink

PR #476, opened 2018-06-30
Fix bug in PipeSnippetStep where outputs are split incorrectly result…

ekkizogloy:
…ing in a mismatch of outputs and commands

The processing of outputs failed to properly split output properly.


ekkizogloy:
Archiving for later.

PR #481, opened 2018-07-17
Feat connect direct

ekkizogloy:


iotiletest/iotile/fixture/device_fixture.py
```diff
@@ -23,8 +24,12 @@ def pytest_generate_tests(metafunc):
         ids = [int(x, 0) for x in metafunc.config.option.tile]
         metafunc.parametrize("tile", ids, scope='session')
 
+    if 'direct' in metafunc.fixturenames:
+        connections = [x for x in metafunc.config.option.device]
+        metafunc.parametrize("direct", connections, scope='session')
+
 @pytest.fixture(scope='module')
-def device(port, device_id):
+def device(port, device_id, direct):
```
timburke: For backwards compatibility of the function signature, this should have a default value of `direct=False`.  I believe this would break otherwise if the user didn't pass `--direct` on the command line.

ekkizogloy: I have tried `direct=False` and `direct=[]` and in both cases the default value gets used rather than the overridden value. I'm not sure exactly what is going on. I did add a default in the option parser though so that should take care of the default case.

transport_plugins/jlink/iotile_transport_jlink/jlink.py
```diff
@@ -299,6 +299,17 @@ def _open_script_interface(self, conn_id, callback):
 
         callback(conn_id, self.id, True, None)
 
+    def _open_streaming_interface(self, conn_id, callback):
```
timburke: Is this required to add now?  The JLink fw and python package does not support streaming so trying to open the streaming interface should raise a `StreamOperationNotSupported` error which should happen already just by not implementing this function.

Adding this function will make that error go away but streaming will still fail since the underlying channel doesn't support it yet.

I would recommend removing this function until we actually add jlink support for streaming.  It will mean that a call to `enable_streaming` in HardwareManager will raise an exception but I would argue that is correct since streaming is not supported.

ekkizogloy: Ok.. I added this function before discussing the issues with streaming... I'll remove it.

PR #487, opened 2018-08-02
Native BLE transport plugin

haroal:
Here it is! The new transport plugin using native BLE thanks to [baBLE](https://github.com/iotile/baBLE).
It basically works like any transport plugin, to use it simply run:
```bash
# If you want it to automatically choose a controller
$ iotile hw --port="ble"
# Or if you want to choose it manually (with X the number in hciX if you run `sudo hcitool dev`)
$ iotile hw --port="ble:X"
```

For now, it only works on **Linux**.

PS: to make the tests passing, although it only works on Linux, I had to modify the root `tox.ini` and the `scripts/test.py` script to skip tests if on Mac or Windows.


transport_plugins/native_ble/iotile_transport_native_ble/tilebus.py
```diff
@@ -0,0 +1,54 @@
+from bable_interface import Characteristic, Service
+import uuid
+
+# Company ID
+ArchManuID = 0x03C0
+
+# GATT table
+BLEService = Service(uuid='1800', handle=0x0001, group_end_handle=0x0005)
```
timburke: Are these handles and `group_end_handle` only used on the virtual interface?  If so that's totally fine but if you key in on the handles for PODs as well that will break in the future as soon as we tweak a ble characteristic on the POD firmware and the handles are reassigned.  In that case I wouldn't fix it now but file an issue to we remember to fix it in the future when we make such a change, which won't be too soon.

transport_plugins/native_ble/iotile_transport_native_ble/device_adapter.py
```diff
@@ -0,0 +1,995 @@
...
+        result = {}
+
+        def on_header_received(value):
+            """Callback function called when a notification has been received with the RPC header response."""
+            result['status'] = value[0]
+            result['length'] = value[3]
+
+            if result['length'] > 0:
+                # Register the payload notification callback
+                self._register_notification_callback(
```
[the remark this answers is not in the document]
haroal: I was wondering the same thing a few days ago and then I forgot to verify. It never happens during my tests but I am not sure enough to tell you that it is impossible. So, I will definitely register the payload callback before and remove it if it is not needed anymore. Would be better for sure :)

transport_plugins/native_ble/iotile_transport_native_ble/tilebus.py
haroal: It is just that currently bable_interface.Service object are not linked to bable_interface.Characteristic: there is no hierarchy because it was quite annoying to serialize/unserialize then. But it is definitely something to change in baBLE: add an `add_characteristic` method to add a characteristic to a service and then, automatically calculate the `group_end_handle`, and the Characteristics `handle`,`value_handle`, `config_handle`.

This information is needed to register the GATT table so I need them only in the virtual interface.

So yes for now, we'll have to keep this data the same as the GATT table registered in the PODs. I'll create an issue

haroal: After thinking about that, we'll always have to keep this updated when modifying the GATT table in the POD firmware, just like we have to in the BLED112 transport plugin (cf [virtual_bled112.py#L50](https://github.com/iotile/coretools/blob/master/transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py#L50)). 

So even if adding the Service/Characteristic hierarchy would be great (cf [issue here](https://github.com/iotile/baBLE/issues/23)), it won't solve this issue.

Issue #488, opened 2018-08-02
iotile-build  - update autobuild_trub_script to include app_info and sensorgraph


x24ling:
Function header should be 
```autobuild_trub_script(file_name, slot_assignments=None, os_info=None, sensorgraph=None, app_info=None)```

PR #493, opened 2018-08-03
Upgrade iotile ship

x24ling:
Addresses the following:
iotile-ship
-VerifyDeviceStep uses shared resources #489
-SyncCloudStep has overwrite (default false) parameter #490 

iotile-build
-autobuild_trub_script(file_name, slot_assignments=None, os_info=None, sensor_graph=None, app_info=None)  #488

### DIFF
--- a/iotilebuild/iotile/build/config/site_scons/trub_script.py
+++ b/iotilebuild/iotile/build/config/site_scons/trub_script.py
@@ -10,9 +10,11 @@ from iotile.core.hw.update import UpdateScript
 from iotile.core.hw.update.records import ReflashTileRecord, ReflashControllerRecord, SetDeviceTagRecord
 from iotile.build.build import ProductResolver
 from iotile.core.utilities.intelhex import IntelHex
+from iotile.sg.compiler import compile_sgf
+from iotile.sg.output_formats.script import format_script
+from iotile.core.hw.update.script import UpdateScript
 
-
-def build_update_script(file_name, slot_assignments, os_info=None):
+def build_update_script(file_name, slot_assignments=None, os_info=None, sensor_graph=None, app_info=None):
     """Build a trub script that loads given firmware into the given slots.
 
     slot_assignments should be a list of tuples in the following form:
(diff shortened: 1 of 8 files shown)


iotilebuild/iotile/build/config/site_scons/trub_script.py
```diff
@@ -27,51 +29,77 @@ def build_update_script(file_name, slot_assignments, os_info=None):
...
-            record = ReflashControllerRecord(bin_data, offset)
-        else:
-            record = ReflashTileRecord(slot_id, bin_data, offset)
-
-        records.append(record)
-
+    
+    #Update sensorgraph
+    app_info = env['APP_INFO']
+    if app_info is not None:
```
timburke: The firmware images for each time should be programmed first in case there is a dependency of the sensorgraph on some particular functionality present in the firmware images.

timburke: This assumes that app_info and sensorgraph will always be set together, even though nothing in the calling function requires that (they are both separate parameters with default values that can be set independently).  If you want to ensure they are set together you'll need add a check in the calling function that throws an exception if `app_info` but no `sensor_graph`.

However, I would guess that it would be better to just allow them both to be set independently and remove the assumption in this function that `app_info is not None` means sensorgraph.  You can use `ENV` to pass whatever additional flags you want to this function.

iotilebuild/iotile/build/config/site_scons/trub_script.py
```diff
@@ -27,51 +29,77 @@ def build_update_script(file_name, slot_assignments, os_info=None):
...
+            else:
+                record = ReflashTileRecord(slot_id, bin_data, offset)
+
+            records.append(record)
+
+    #Update App and OS Tag
     os_info = env['OS_INFO']
     if os_info is not None:
         os_tag, os_version = os_info
         records.append(SetDeviceTagRecord(os_tag=os_tag, os_version=os_version))
```
timburke: You can do this with a single call to `SetDeviceTagRecord` with all 4 parameters set or None if you don't want to specify the app or os part of the tag.

https://github.com/iotile/coretools/blob/85e7caa3a31f1b59c6fbe5b7ee7f47d731399d47/iotilecore/iotile/core/hw/update/records/set_version.py#L22

iotilebuild/iotile/build/config/site_scons/trub_script.py
x24ling: Good point, i'm going to allow them to be set independently.

PR #494, opened 2018-08-07
Feat v2 advpkt

ekkizogloy:
First pass at the v2 advertising packet support in the bled112 adapter and virtual device. The implementation in the virtual device is very limited with hardcoded values.


ekkizogloy:
Closed to fix tests.

PR #500, opened 2018-08-17
Feat v2 advpkt

ekkizogloy:
Added initial implementation of coretools to support advertising packets v2. 
Scan outputs both version 1 and version 2 packets with an indicator showing version.


transport_plugins/bled112/RELEASE.md
```diff
@@ -2,6 +2,10 @@
 
 All major changes in each released version of the bled112 transport plugin are listed here.
 
+## 1.8.0
```
timburke: Let's  not release this just yet.  This is a core code path that affects all coretools installations including those on nodes and gateways.  If you want to do a prerelease to make this easier to install and test, go for that otherwise, since no one else has v2 advertising packets, you can test locally with a `pip install -e <path to folder>`

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -504,15 +505,142 @@ def _handle_event(self, event):
         else:
             self._logger.warn('Unhandled BLE event: ' + str(event))
 
+
     def _parse_scan_response(self, response):
         """Parse the IOTile specific data structures in the BLE advertisement packets and add the device to our list of scanned devices
         """
+        self._logger.debug("   call: _parse_scan_response({0})".format(response))
```
timburke: These debug logging statements seem like they're leftover from initial development learning what functions were calling what and should be removed before.  In particular:

- these statements seem one or two levels two chatty.  This one could be called hundred of times per second if there is a single non-iotile ble device around that advertises frequently
- logging statements should not use `.format` to format the log message.  I know that generally `.format` is preferred for string replacement but the `logging` module is an exception to that rule even in python 3.  The reason is that if you call .format then the string replacement happens immediately every time `debug` is called which is potentially expensive, even if the user hasn't enabled that level of logging.  Just passing the arguments to `log` instead allows the logging module to see if log statement is sent anywhere before doing any argument conversion.  See:
https://docs.python.org/3/library/logging.html#logging.Logger.debug

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -504,15 +505,142 @@ def _handle_event(self, event):
...
+        # Parse advertisement and determine if it matches V1 or V2.
 
         payload = response.payload
         length = len(payload) - 10
 
         if length < 0:
-            return  # FIXME: Log an error here
+            self._logger.error("BLE Advertising packet payload is empty.")
```
timburke: I actually think that old `FIXME` is wrong and we should not log an error here.  This will log an error for all non-iotile devices that don't happen to have a lot of scan response data.

transport_plugins/bled112/iotile_transport_bled112/bled112_cmd.py
```diff
@@ -16,6 +17,51 @@
 
 BGAPIPacket = namedtuple("BGAPIPacket", ["is_event", "command_class", "command", "payload"])
 
+
+
+#DEBUG Print helpers
```
timburke: These initial development aids should be removed before merging.

transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
```diff
@@ -55,9 +56,15 @@ class BLED112VirtualInterface(VirtualIOTileInterface):
     HighspeedHandle = 21
     TracingHandle = 23
 
+    AdvertisingPacketVer = 1
+    json_args = None
```
timburke: This global variable doesn't seem to do anything.

transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
```diff
@@ -55,9 +56,15 @@ class BLED112VirtualInterface(VirtualIOTileInterface):
...
+
     def __init__(self, args):
         super(BLED112VirtualInterface, self).__init__()
 
+        if args:
```
timburke: Args is always passed as a dict (empty if no arguments) and will not ever be `None` (if it were None, line 69 would raise an exception).  It will not always be json though so the variable name `json_args` is misleading.  It is only json in one particular case where you invoke this class from the built-in `virtual-device` command line wrapper, but you're free to create this class programmatically yourself as well.

transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
```diff
@@ -193,22 +196,56 @@ def _queue_reports(self, *reports):
...
+            reboots_lo = (reboots & 0x00FFFF)
+
+            data1 = struct.pack("<BBHL", 27, 0x16, 0x03C0, self.device.iotile_id)
+            data2 = struct.pack("<HBBLBB", reboots_lo, reboots_hi, flags, timestamp, voltage, OTHER)
+            data3 = struct.pack("<HLL", bcast_stream, bcast_value, mac)
+
+            return ble_flags + data1 + data2 + data3
+
+        else:
+            self._logger.error("Invalid Advertising Packet Version")
```
timburke: You should raise an exception here rather than that returning `None`.  The callers of this function probably cannot handle a `None` return value so it will just mask the source of the error.

transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
```diff
@@ -55,9 +56,15 @@ class BLED112VirtualInterface(VirtualIOTileInterface):
...
 
+    AdvertisingPacketVer = 1
+    json_args = None
+
     def __init__(self, args):
         super(BLED112VirtualInterface, self).__init__()
 
+        if args:
+            json_args = args
+
```
timburke: Actually, I see that you are deferring the check on what is contained in `args` until you use it.  A better way is to capture the values you need from args with their defaults if not present in __init__, validate them to be valid and then you know their values are valid in the rest of the methods:

```python
advertising_version = args.get('advertising_version', 1)
if advertising_version not in (1, 2):
    raise ArgumentError("Invalid advertising version specified in args", supported=(1, 2), found=advertising_version)
```

With this method you don't need to save json_args around and people can look in a single place (this `__init__` function to see what you expect to find in args.

Also, AdvertisingPacketVer should conform to normal python naming convention of snake case since it's a normal class instance variable.

transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
```diff
@@ -193,22 +196,56 @@ def _queue_reports(self, *reports):
...
+            data3 = struct.pack("<HLL", bcast_stream, bcast_value, mac)
+
+            return ble_flags + data1 + data2 + data3
+
+        else:
+            self._logger.error("Invalid Advertising Packet Version")
+            return None
 
     def _scan_response(self):
+        if self.AdvertisingPacketVer > 1:
```
timburke: Same here, you should throw an exception rather than eating the invalid condition.

transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
```diff
@@ -108,6 +115,46 @@ def __init__(self, args):
...
+        if self.virtual_info['advertising_version'] not in (1, 2):
+            raise ArgumentError("Invalid advertising version specified in args",
+            supported=(1, 2), found=self.virtual_info['advertising_version'])
+
+        self.virtual_info['reboot_count'] = int(args.get('reboot_count', 1), 0)
+        if self.virtual_info['reboot_count'] <= 0:
+            raise ArgumentError("Reboot count must be greater than 0.",
+            supported="> 0", found=self.virtual_info['reboot_count'])
+
+        self.virtual_info['broadcast_stream'] = int(args.get('broadcast_stream', 0xDEAD), 0)
```
timburke: `broadcast_stream` and `broadcast_value` should come from the virtual device that is being served rather than the virtual interface.  By default nothing is broadcast but if the virtual_device broadcasts a value it is saved in the virtual interface and used to update the advertising packet.  

See: https://github.com/iotile/coretools/blob/master/transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py#L187

for how the `broadcast_reading` is naturally updated when a device that is being served streams a broadcast reading.

transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
```diff
@@ -108,6 +115,46 @@ def __init__(self, args):
...
+            supported="0 - 0xFFFFFFFF", found=self.virtual_info['mac_value'])
+
+        self.virtual_info['battery_voltage'] = float(args.get('battery_voltage',"3.14159"))
+        if self.virtual_info['battery_voltage'] < 0 or self.virtual_info['battery_voltage'] > 7.9:
+            raise ArgumentError("Battery voltage is invalid",
+            supported="0 - 7.9", found=self.virtual_info['battery_voltage'])
+
+
+    def init_logger(self):
+        formatter = logging.Formatter('%(asctime)s.%(msecs)03d %(levelname).3s %(name)s %(message)s', '%y-%m-%d %H:%M:%S')
```
timburke: You usually don't want to setup a logging StreamHandler this deep into the class hierarchy.  The default logging behavior is to propagate log messages hierarchically up the tree of loggers based on the dots in their name until it reaches one that has a handler.  Putting a StreamHandler on this logger will usually cause duplicate log messages when the user also turns on a global log handler.  

If the goal was to avoid a warning message saying 'no handler attached to logger' or something to that effect, the best way to resolve that is to attach a `NullHandler` to the logger.

For setting the log level, I would default to keeping it unspecified by default since its up to the user to decide what they want to see but if you do set it, it should use constants like `logging.ERROR` or `logging.DEBUG` instead of bare numbers for clarity.

transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
```diff
@@ -182,35 +229,80 @@ def _queue_reports(self, *reports):
...
+            subsecond_cnt = 0xF #TODO
+
+            reserved = 0
+            with self._broadcast_lock:
+                if self._broadcast_reading is not None:
+                    bcast_stream = self._broadcast_reading.stream
+                    bcast_value = self._broadcast_reading.value
+                    timestamp = calendar.timegm(time.gmtime())
+                else:
+                    #XXX: OK?
```
timburke: I would recommend not doing this.  It means that every access point we load this code on will broadcast a 'deadbeef' broadcast value.  If you're looking for a sentinel value to indicate no broadcast value, we support the following:

stream: 0xFFFF
timestamp: 0xFFFFFFFF
value: 0xFFFFFFFF

The only important part is the reserved invalid stream id `0xFFFF` but the v1 advertisement sets everything to FF so might as well be consistent.

transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
```diff
@@ -501,3 +593,17 @@ def _send_trace(self, chunk=None):
                 traceback.print_exc()
                 print("*** END EXCEPTION ***")
                 self._audit('ErrorStreamingTrace')  # If there was an error, stop streaming but don't choke
+
+
+    #FIXME Need method to set these values
```
timburke: I don't think these methods are needed since the broadcast value is already updated when the virtual device streams a broadcast reading.

transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py
timburke: Note that the rest of `bled112` transport plugin is a bad guide for good logging behavior.  It has the worst logging behavior of any package (because I wrote it before coretools was large enough for good logging behavior to matter and didn't know any better...)

timburke: If you're looking to make sure your logging messages actually show up, the place to make the change is here:
https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/scripts/virtualdev_script.py#L49

You just need to add the `StreamHandler` on the root logger.

Issue #503, opened 2018-08-20
HardwareManagerResource missing connect_direct as RESOURCE_ARG_SCHEMA


x24ling:
Currently can only connect with uuid, should allow connect_direct as an option.

PR #505, opened 2018-08-21
Add custom steps before build, HardwareResrouce allow connect_direct

### LINKED ISSUE #504: iotile-ship - autobuild_shiparchive unable to create .ship archives if a custom Step is necessary
For the instance of building scripts for the pod_1m, which relies on a custom ActionStep, CalibratePOD1MStep, it fails 
```
 File "c:\users\chris\iotile\coretools\iotilebuild\iotile\build\config\scons-local-3.0.1\SCons\Action.py", line 1197, in execute
    result = self.execfunction(target=target, source=rsources, env=env)
  File "c:\users\chris\iotile\coretools\iotileship\iotile\ship\autobuild\ship_file.py", line 107, in create_shipfile
    resman.add_recipe_folder(source_dir, whitelist=[os.path.basename(str(source[0]))])
  File "c:\users\chris\iotile\coretools\iotileship\iotile\ship\recipe_manager.py", line 79, in add_recipe_folder
    recipe = RecipeObject.FromFile(os.path.join(recipe_folder, yaml_file), self._recipe_actions, self._recipe_resources)
  File "c:\users\chris\iotile\coretools\iotileship\iotile\ship\recipe.py", line 215, in FromFile
    action=action_name, step=i + 1, path=path)
UnknownRecipeActionType: UnknownRecipeActionType: Unknown step specified in recipe
Additional Information:
action: CalibratePOD1MStep
path: build\none\calibrate_pod1m\calibrate_pod1m.yaml
step: 3
```

The reason for this is that autobuild_shiparchive calls the following functions
```
resman = RecipeManager()
resman.add_recipe_folder(source_dir, whitelist=[os.path.basename(str(source[0]))])
recipe = resman.get_recipe(recipe_name)
```

Which relies on RecipeManager reading from entry_points to determine the list of valid RecipeActions, but this is a problem if this package has not been installed yet, which is the case when we're building the package.

x24ling:
Fixes #504  and #503

### DIFF
--- a/iotileship/iotile/ship/autobuild/ship_file.py
+++ b/iotileship/iotile/ship/autobuild/ship_file.py
@@ -3,14 +3,14 @@
 This file is only meant to be used in conjunction with iotile-build and
 should not be imported independently.  It is only for use inside SConstruct
 files."""
-
 import os
+import imp
+import inspect
 from SCons.Script import Environment, Copy, Action
 from iotile.build.build import ProductResolver, ArchitectureGroup
 from iotile.build.utilities import render_template_inplace
 from iotile.core.exceptions import BuildError
 from ..recipe_manager import RecipeManager
-
 
 def autobuild_shiparchive(src_file):
     """Create a ship file archive containing a yaml_file and its dependencies.
@@ -35,6 +35,19 @@ def autobuild_shiparchive(src_file):
     family = ArchitectureGroup('module_settings.json')
     target = family.platform_independent_target()
     resolver = ProductResolver.Create()
+
+    #Parse through build_step products to see what needs to imported
+    build_steps = family.tile.build_steps()
+    custom_steps = []
+    for build_step in build_steps:
+        full_file_name = build_step.split(":")[0]
+        basename = os.path.splitext(os.path.basename(full_file_name))[0]
+        folder = os.path.dirname(full_file_name)
+
+        fileobj, pathname, description = imp.find_module(basename, [folder])
+        mod = imp.load_module(basename, fileobj, pathname, description)
+        custom_steps += filter(lambda c: c[1].__module__==basename, inspect.getmembers(mod, inspect.isclass))
+    env['CUSTOM_STEPS'] = custom_steps
 
     env["RESOLVER"] = resolver
 
@@ -104,6 +117,8 @@ def create_shipfile(target, source, env):
     recipe_name = os.path.basename(str(source[0]))[:-5]
 
     resman = RecipeManager()
+
+    resman.add_recipe_actions(env['CUSTOM_STEPS'])
     resman.add_recipe_folder(source_dir, whitelist=[os.path.basename(str(source[0]))])
     recipe = resman.get_recipe(recipe_name)
 
--- a/iotileship/iotile/ship/autobuild/ship_file.py
+++ b/iotileship/iotile/ship/autobuild/ship_file.py
@@ -46,7 +46,8 @@ def autobuild_shiparchive(src_file):
 
         fileobj, pathname, description = imp.find_module(basename, [folder])
         mod = imp.load_module(basename, fileobj, pathname, description)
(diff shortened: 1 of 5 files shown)


x24ling:
I have also verified that the pod1m calibration recipes get properly built with this package locally installed on my machine.

iotileship/iotile/ship/autobuild/ship_file.py
```diff
@@ -36,6 +36,19 @@ def autobuild_shiparchive(src_file):
...
+    build_steps = family.tile.build_steps()
+    custom_steps = []
+    for build_step in build_steps:
+        full_file_name = build_step.split(":")[0]
+        basename = os.path.splitext(os.path.basename(full_file_name))[0]
+        folder = os.path.dirname(full_file_name)
+
+        fileobj, pathname, description = imp.find_module(basename, [folder])
+        mod = imp.load_module(basename, fileobj, pathname, description)
+        custom_steps += filter(lambda c: c[1].__module__==basename, inspect.getmembers(mod, inspect.isclass))
```
timburke: Two comments on this line:

1. It is better to use a list comprehension than a `filter` statement because it's more clear what is going on.  You could rewrite this line as:
```python
custom_steps += [x for x in inspect.getmembers(mod, inspect.isclass) if x[1].__module__ == basename)]
```

2. Unlike the proxy examples in HardwareManager, you know exactly what the build step class name is since it's the part of `build_step` after the colon.  So, you can just directly get it from the module without needing the `inspect` part:

```python
full_file_name, class_name = build_step.split(":")
custom_step = getattr(mod, class_name)
```

iotileship/iotile/ship/recipe_manager.py
```diff
@@ -86,6 +86,17 @@ def add_recipe_folder(self, recipe_folder, whitelist=None):
...
 
+    def add_recipe_actions(self, recipe_actions):
+        """Add additional valid recipe actions to RecipeManager
+
+        args:
+            recipe_actions (list): List of tuples. First value of tuple is the classname, 
+                second value of tuple is RecipeAction Object
+
+        """
+        for recipe_action in recipe_actions:
```
timburke: For clarity, it's better to break out the parts of `recipe_action` explicitly here using a destructuring assignment:

```python
for action_name, action in recipe_actions:
    self._recipe_actions[action_name] = action
```

iotileship/iotile/ship/resources/hardware_manager.py
```diff
@@ -10,7 +10,7 @@
 RESOURCE_ARG_SCHEMA = DictionaryVerifier(desc="hardware_manager arguments")
 RESOURCE_ARG_SCHEMA.add_optional("port", StringVerifier("the port string to use to connect to devices"))
 RESOURCE_ARG_SCHEMA.add_optional("connect", OptionsVerifier(StringVerifier(), IntVerifier(), desc="the uuid of the device to connect to"))
-
+RESOURCE_ARG_SCHEMA.add_optional("connect_direct", OptionsVerifier(StringVerifier(), IntVerifier(), desc="the connection string of the device to connect to"))
```
timburke: I think connect_direct is always a string.  Even when the value is a single integer, the type of `connection_string` is always a string unless you know of a bug where that isn't the case.

iotileship/iotile/ship/autobuild/ship_file.py
x24ling: oh wow, good point, fixed

iotileship/iotile/ship/recipe_manager.py
x24ling: done

iotileship/iotile/ship/resources/hardware_manager.py
x24ling: that is correct, it should only be a string when connect_direct args is used.

PR #507, opened 2018-09-08
Bug fix to prevent acknowledgement of stream ids > 255.

ekkizogloy:


iotile_ext_cloud/iotile/cloud/apps/cloud_uploader.py
```diff
@@ -91,13 +91,14 @@ def _wait_streamers_finished(self, timeout=60*10.0):
         raise HardwareError("Device took too long to stream data", timeout_seconds=timeout)
 
     def _ack_streamer(self, index, value):
-        error, = self._con.rpc(0x20, 0x0f, index, 0, value, arg_format="HHL", result_format="L")
+        if (index <= 0xFF):
```
timburke: This function should raise an `ArgumentError` if called with an invalid streamer id rather than just silently ignoring the command.  That will make sure users know they are using it incorrectly and that it did not actually acknowledge a streamer for them.

iotile_ext_cloud/iotile/cloud/apps/cloud_uploader.py
```diff
@@ -147,8 +148,9 @@ def download(self, trigger=None, acknowledge=True):
                 index = ack['index']
                 last_id = ack['last_id']
 
-                self.logger.info("Acknowledging highest ID %d for streamer %d", last_id, index)
-                self._ack_streamer(index, last_id)
+                if (index <= 0xFF):
```
timburke: idiomatic python, unlike C does not require parenthesis around conditionals.

Issue #508, opened 2018-09-19
Cloud Uploader App Does Not Configure Longer Report Lengths


timburke:
Before calling `enable_streaming` the `cloud_uploader` app needs to send an RPC to tell the POD to use longer report chunks so that large reports are not broken up into multiple pieces.

timburke:
The needed call is right before this line:

https://github.com/iotile/coretools/blob/master/iotile_ext_cloud/iotile/cloud/apps/cloud_uploader.py#L161

The required RPCs that need to be added are:
- `0x0A06`: query_stream_parameters()
- `0x0A05`: config_stream_parameters()

`query_stream_parameters` takes no arguments and returns a packed `struct` with format `LBB`:
- the first returned value is a 32-bit number that says the currently configured maximum report length in bytes
- the second parameter is a bitmap of supported compression types
- the third parameter is the currently selected compression type

Both the second and third parameters are unuses.

`config_stream_parameters` takes an `LB` sctruct with:
- the first parameter being the desired maximum packet size
- the second parameter being the desired compression type.  The second parameter should always be 0.

The first parameter should be set to `2*1024*1024` for 2MB maximum reports.

timburke:
This was closed by PR #509

PR #509, opened 2018-09-20
Set max report size before downloading reports

ekkizogloy:
Set max_report size to maximum to essentially disable report chopping.

### DIFF
--- a/iotile_ext_cloud/iotile/cloud/apps/cloud_uploader.py
+++ b/iotile_ext_cloud/iotile/cloud/apps/cloud_uploader.py
@@ -158,6 +158,8 @@ class CloudUploader(IOTileApp):
         else:
             self.logger.info("Not acknowledging readings from cloud per user request")
 
+        # Configure Downloader to not break up the report
+        self._con.config_streaming(max_packet=0xFFFFFFF0)
         self._hw.enable_streaming()
 
         if trigger is not None:
--- a/iotile_ext_cloud/RELEASE.md
+++ b/iotile_ext_cloud/RELEASE.md
@@ -2,6 +2,10 @@
 
 All major changes in each released version of the iotile-ext-cloud plugin are listed here.
 
+## 0.5.2
+
+- Set max report size to max in cloud uploader so that chopping of reports is disabled. 
+
 ## 0.5.1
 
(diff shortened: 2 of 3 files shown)


iotile_ext_cloud/iotile/cloud/apps/cloud_uploader.py
```diff
@@ -158,6 +158,8 @@ def download(self, trigger=None, acknowledge=True):
         else:
             self.logger.info("Not acknowledging readings from cloud per user request")
 
+        # Configure Downloader to not break up the report
+        self._con.config_streaming(max_packet=0xFFFFFFF0)
```
timburke: This app is in a coretools package which means in needs to not have any dependencies on non-coretools packages.  The `config_streaming` method is provided by `iotile-support-con-nrf52832`.  That means that adding this call will make the CloudUploader break in the following cases:
- user does not have `iotile-support-con-nrf52832` installed
- user is not talking to an nRF52832 based POD but still wants to upload data from it (such as a node or other virtual device).  

The goal with CoreTools apps such as CloudUploader is to provide global implementations of specific use cases (in this case uploading streamer reports to the cloud) that do not rely on functionality provided by the specific proxies of any given device but instead directly call the relevant RPCs directly and work with any device that implements those RPCs.

That's why there are no other reference to any con-nrf52832 specific functionality in this class; note that all of the other RPCs that are invokes are done using the direct `self._con.rpc` method which all devices provide, not assuming that there will be a proxy module installed that provides a convenience python wrapper, see the `basic=True` parameter here:
https://github.com/iotile/coretools/pull/509/files#diff-64ff6b7ed94c181141c2d873c3fc7854L43

The required rpcs that would nee to be added to CloudUploader are documented here:
https://github.com/iotile/coretools/issues/508#issuecomment-422848894

Also, not all devices support configuring the streamer report length so this method needs to gracefully handle when the RPC is not implemented (and raises a HardwareError).

ekkizogloy: Thanks for that feedback. Actually... cloud_uploader.py already implements set_report_size and handles the errors as needed.... I just needed to call that method.

Issue #445, opened 2018-04-30
Add `watch` and `watch_reports` to HardwareManager


timburke:
There should be a way to just have `iotile` print out newly received scan information with a filter and to print out each new report as it comes in for debugging purposes.

### Example Usage

Should print every newly broadcast reading from a specific device
`iotile hw watch --uuid 0x100 --only-readings`

Should block until Ctrl-C and print information about every new report received from a device as it comes in.
`iotile hw watch_reports --include-streams=[0x5001, 0x5002]`

timburke:
A variable of this has been implemented for watching broadcast reports that can provide a template for watching realtime streams.  It uses the `asciimatics` package to provide a console based ui that lets you update the values for each stream in place.

https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/hw/hwmanager.py#L438

timburke:
The remaining work to be done here is to implement `watch_reports` which is invokable after connecting to a device and supports the following arguments:

- whitelist_streams: list of int, if given will only show those streams
- blacklist_streams: list of int, if given will show all streams other than those specific

By default all streams from the device will be shown.

mattrunchey:
@timburke Is the functionality described in this issue implemented with #510 ?

timburke:
Yes it is 👍

PR #510, opened 2018-09-21
Feat watch 445

mattrunchey:
Addresses #445 and cleans up some of the code in affected files WRT pylint. 

Testing done:
* Executed function against multiple virtual devices (connect/disconnect, enable_streaming, etc)
* Executed function against real devices and verified that report results were showing up and changing over time
* Ran tests locally in the iotilecore/tests folder (this prompted the do_test changes) against python3 

Other review notes:
* travis-ci tests passing with @ekkizogloy 's fix on master

### DIFF
--- a/iotilecore/iotile/core/hw/hwmanager.py
+++ b/iotilecore/iotile/core/hw/hwmanager.py
@@ -476,13 +476,13 @@ class HardwareManager:
             blacklist = set(blacklist)
 
         def _title(_items):
-            return "Watching Broadcast Reports (Ctrl-C to Stop)"
+            return ["Watching Broadcast Reports (Ctrl-C to Stop)"]
 
         def _poll():
             results = [x for x in self.iter_broadcast_reports(blocking=False)]
             return results
 
-        def _text(item, screen):
+        def _text(item):
             fmt_uuid = "%08X" % item.origin
             fmt_uuid = fmt_uuid[:4] + '-' + fmt_uuid[4:]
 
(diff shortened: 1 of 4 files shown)


iotilecore/iotile/core/hw/hwmanager.py
```diff
@@ -34,7 +39,7 @@
 
 
 @context("HardwareManager")
-class HardwareManager(object):
+class HardwareManager:
```
timburke: Quick question.  My understanding was that in python 2.7, you still needed to add the `(object)` explicitly to make this a new-style class and that was only unnecessary on python 3+?  Am I missing something?

mattrunchey: I may be misreading, but in python's docs they claim new-style classes have been integrated with 2.7 https://www.python.org/doc/newstyle/

timburke: Yeah it looks like you're right, til:

```
Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> class test_class: pass
...
>>> obj = test_class()
>>> isinstance(obj, object)
True
```

mattrunchey: It's confusing because they say they haven't updated the python docs, I think it's probably safer to keep (object) in place. I'll add them in, more important than fixing a python3 linting error.

PR #521, opened 2018-10-02
Fix 520

mattrunchey:
Conditionals now properly exit before hitting an IndexError. 

Separated functional change from style change, so two commits.

Testing done:

Had some printout code to see that the conditional was dropping to the "else". Caught plenty of devices that weren't Arch devices. Was unable to actually replicate or find another device that was advertising with 0 or 1 packets, unfortunately, but the behavior will be handled now.


mattrunchey:
Ah, good call. I'll add a commit bumping the version and release notes before merging.

PR #523, opened 2018-10-03
Enable iotile-build to work with python3

mattrunchey:
Basically, after a few import fixes and dict_keys, and changing the wheel
to be non hard coded to py2, "iotile build" successfully executed against a tile_gpio build.

Testing done: 
```
(ota_dev) Matts-Macbook-Pro:firm_gpio mrunchey$ iotile build
```
Artifacts created (.elf, .whl).


iotilecore/iotile/core/dev/iotileobj.py
```diff
@@ -239,7 +240,13 @@ def _load_settings(self, info):
 
         # Setup our support package information
         self.support_distribution = "iotile_support_{0}_{1}".format(self.short_name, self.parsed_version.major)
-        self.support_wheel = "{0}-{1}-py2-none-any.whl".format(self.support_distribution, self.parsed_version.pep440_string())
+        if sys.version_info[0] >= 3:
```
timburke: We should take this opportunity to add support for universal wheels if the module specifies that it supports it.  As implemented, we won't be able to release python 3 wheels from our CI server because it builds with python 2 (and vice-versa if we upgrade the CI server to python 3.

Just like we support `python_depends`, we should also support a `python_universal` flag, see:
https://github.com/iotile/coretools/pull/523/files#diff-f3d1c5503b35ff237b2009dccc2d99a3L196

If `python_universal` is False or not present, we default to the creating the wheel based on the python interpreter being used.  If `python_universal` is set we expect a `.py2.py3` universal wheel

timburke: There's an additional issue that gets added into the mix which is that the wheels of dependencies are also copied into `build/output/python` so that `build/output` contains everything needed to install and use this module (including all of the required python dependencies.

However, there is now an issue because if a dependency was built with python 2 and released (most/all currently are) then this code will look for the `.py3` wheel and not find it since it is `.py2`.

So, we should also add checks in the build process to print friendly build messages if you are building on python 3 and one of your dependencies was built for python 2:

See:
https://github.com/iotile/coretools/blob/master/iotilebuild/iotile/build/config/site_scons/pythondist.py#L91
https://github.com/iotile/coretools/blob/master/iotilebuild/iotile/build/config/site_scons/dependencies.py#L61

mattrunchey: Got it. Will make changes to incorporate this and update!

mattrunchey: also dependency checking for if py3 is build, that the python_depends are compatible (probably good to do it for py2 as well). Might need a better test bed, tile_gpio doesn't seem to have python dependencies.

timburke: I'd recommend adding an end-to-end test of the building process.

There's an example of a test component here:
https://github.com/iotile/coretools/tree/master/iotilebuild/test/test_build/component

You would need to build 2 or 3 such combinations of components with:
- the component under test (with a module_settings.json that specifies its dependencies)
- its dependencies installed in `build/deps` inside the component folder.  This is what `iotile depends update` does, just blatantly copy the `build/output` folder from the dependency into `build/deps` for the component that requires it.  All `iotile build` related things that refer to a dependency just mean the appropriate folder in `build/deps`

You can play around with the module_settings.json file in the `deps/<dependency name>` folder to add/remove `python_universal` and simulate a python 2 / python 3 build by putting a `py2` or `py3` .whl file (can have any contents or be empty) in the `deps/<dependency name>` folder.

timburke: Actually there are much better examples here:
https://github.com/iotile/coretools/tree/master/iotilebuild/test/test_iotilebuild

timburke: https://github.com/iotile/coretools/blob/master/iotilebuild/test/test_iotilebuild/test_scons.py#L75

timburke: @mattrunchey Let me know if you want a quick rundown of how `iotile-build` expects the `deps` folder to structured so that it can find the dependency correctly

mattrunchey: Changes should be addressed with the latest changes that were just force pushed to this commit. two new test directories and 4 new tests

mattrunchey:
Putting an update comment, changes that were requested were addressed @timburke

iotilebuild/iotile/build/doc/dependencies.py
```diff
@@ -51,7 +51,7 @@ def _get_dependencies(self, component):
         with open(module_settings, "r") as f:
             settings = json.load(f)
 
-        mod = settings['modules'].keys()[0]
+        mod = list(settings['modules'].keys())[0]
```
timburke: You can actually just call `list(settings['modules'])` here, the `.keys()` is redundant:
http://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list

mattrunchey:
Closing notes:

- Ran "iotile build" against lib_controller, on python3 -- py3 artifacts actually got built. Also attempted with the universal flag, and saw that the correct wheel got built. Output snippet

```
Copy("build/output/python/iotile_support_lib_controller_3-3.7.3-py2.py3-none-any.whl", "build/python/dist/iotile_support_lib_controller_3-3.7.3-py2.py3-none-any.whl")
Copy("build/output/python/iotile_support_lib_controller_3-3.7.3.tar.gz", "build/python/dist/iotile_support_lib_controller_3-3.7.3.tar.gz")
```

mattrunchey:
I'll leave the branch open for a bit, just in case there are bugs that weren't caught in testing.

PR #531, opened 2018-10-19
Issue 530

### LINKED ISSUE #530: Add device_slug to scan payload
I only find myself having to find a HEX to INT converter to convert my device slug into a UUID so I can find it in the **scan** results.

Does it make sense to add 

```
"slug": "d--......"  (or "device_slug" or "device_id", or "s/n")
```

to the scan payload results?

dkarchmer:
Closes #530 
I also changed the version number, which seems was not changed on the last set of changes. This will be ready to release, if we want to


dkarchmer:
@timburke didn't seem we were testing scanning before, and not simple so I just added the test for the function that creates the slug. Let me know if not enough

PR #534, opened 2018-10-26
Feat ota app

mattrunchey:
OTA app. Typical usage is a one-liner when connected to a device and in the app:

check_and_update

This will check the cloud for a matching deployment that is not marked as complete, apply that update (if applicable), and report to the cloud that it has applied the update (or failed, if so). In the background, this app leverages the device_updater app to do the actual script application.

One thing that might need to get done in addition making sure that we prevent the application of "bad" updates... that is, if the application fails for a device, perhaps the cloud should stop serving the update to other devices until it's verified?


mattrunchey:
Note : these tests won't pass until the python_iotile_cloud changes get merged, because they use the new api endpoints.

iotile_ext_cloud/iotile/cloud/apps/ota_updater.py
```diff
@@ -0,0 +1,210 @@
...
+           "==": operator.ne,
+           "eq": operator.ne
+          }
+
+
+def _download_ota_script(script_url):
+    """Download the script from the cloud service and store to temporary file location"""
+
+    try:
+        urllib.request.urlretrieve(script_url, "temp.trub")
```
timburke: See `requests` package comment.

iotile_ext_cloud/iotile/cloud/apps/ota_updater.py
```diff
@@ -0,0 +1,210 @@
...
+           ">": operator.le,
+           "gt": operator.le,
+           "<": operator.ge,
+           "lt": operator.ge,
+           "==": operator.ne,
+           "eq": operator.ne
+          }
+
+
+def _download_ota_script(script_url):
```
timburke: I would recommend not downloading to disk but just fetching in memory and then passing to `run_script` in `UpdaterApp` which directly takes an `UpdateScript` object:
https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/hw/app/updater_app.py#L139

The `requests` package has a good http `get` routine that will just give you the binary data in memory.

iotile_ext_cloud/iotile/cloud/apps/ota_updater.py
```diff
@@ -0,0 +1,210 @@
...
+        return device_id_to_slug(device_id)
+
+    @docannotate
+    def check_and_update(self):
+        """
+        Checks the cloud for an available OTA update and applies it to the device.
+        Informs the cloud of the attempt afterwards
+        Gives some diagnostic info to the user as process goes
+
+        :return: None
```
timburke: `typedargs` `@docannotate` only understands google style docstrings, not raw rst formatted docstrings so this would need to be formatted as:

```
Returns:
    type: description
```

for `docannotate` to pick it up but the default return value without specifying anything in `None` so the `Returns` block can be omitted.

iotile_ext_cloud/iotile/cloud/apps/ota_updater.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+
+        script = self.check_cloud_for_script()
+        if not script:
+            print("no OTA pending")
+            return
+
+        print("Applying deployment ", script[0])
```
timburke: Similar to Luke's comment, we try to use `iprint` rather than bare print statements when providing status updates that are primarily relevant to interactive invocation of a method from the `iotile` tool.  `iprint` just calls `print` when run from the `iotile` tool but hides the output when run from a background script.  

This is because we tend to end up (by design) reusing this functionality in many other scripts and bare print statements can't be turned off easily in those contexts, causing unwanted noise on stdout.

iotile_ext_cloud/iotile/cloud/apps/ota_updater.py
```diff
@@ -0,0 +1,210 @@
...
+        print("Applying script")
+        status = self._apply_ota_update(filename=filename)
+
+        if status is False:
+            print("OTA update application failed")
+
+        print("Informing cloud of OTA status")
+        try:
+            self._inform_cloud(script[0], self.dev_slug, status)
+        except Exception as e:
```
timburke: `annotated` methods designed to be invoked from the `iotile` tool should generally raise exceptions to indicate errors just like normal methods.  

The reasoning is because these methods are designed to be invoked both from the `iotile` tool as well as from scripts.  If the methods don't raise exceptions they can't be safely integrated into scripts because there's no way to know if they succeeded.  So, instead, we made the `iotile` tool handle exceptions raised from methods it calls gracefully and print them as error messages.  That way, you can just raise a normal exception when there is an issue and it will be printed without a stacktrace in the `iotile` tool and raised normally in a script.

The overall goal is to design a method that can be used interactively from the `iotile` shell without compromising its ability to be used as a normal python api from a script.

iotile_ext_cloud/iotile/cloud/apps/ota_updater.py
```diff
@@ -0,0 +1,210 @@
...
+        return deployment_id, script_url
+
+    def _apply_ota_update(self, filename="temp.trub"):
+        """"Attempt to apply script to device using the device_updater app"""
+
+        updater = self._hw.app(name='device_updater')
+        try:
+            updater.load_script(filename, confirm=False)
+            return True
+        except Exception as e:
```
timburke: See above comment on letting exceptions propagate through to the `iotile` tool

iotilecore/iotile/core/dev/semver.py
```diff
@@ -342,8 +375,10 @@ def FromString(cls, range_string):
             disjuncts = [[conj]]
 
         #Check for ^X.Y.Z
-        elif range_string[0] == '^':
+        elif range_string[0] in('^', '>='):
```
timburke: I don't think the semantics of the `>=` operator are correct here.  It looks like it behaves identically to `^` but >= should not have an upper bound, whereas `^` should.

mattrunchey: I was basing this on the documented comment --
>         ^X.Y.Z - matches all versions with the same leading nonzero digit
>           greater than or equal the given range.

timburke: The documented behavior is correct for the ^ operator.  It should do "semantic versioning" compatible matching, which is any version with the same major version number and minor+patch >= the given version.  

However `>=` is not the same operation.  `>=` should just match any version number greater than or equal to the specified version with no upper bound.  So, for example:

`^1.2.0` would match (1.2.0, 1.3.0, but not 2.0.0) whereas
`>=1.2.0` would match (1.2.0, 1.3.0, and 2.0.0)

mattrunchey: Ah, I see. It turns out I didn't actually need to keep this change in since I used the verisioning a bit differently, so I took it out since my code wasn't exercising the path / no use case for it yet.

mattrunchey:
@timburke Addressed all of the changes you recommended with comments where needed. Behavior now consistent across platforms AND python versions < 2.7.15 :)

Some notes for those who may venture here again in the future:

 - Python 2.7.15 implemented __ne__ in total_ordering(). Our semver didn't have that in it, so our tests that actually leveraged the operator prior to that failed (this PR seems to be the first automatic test that exercises semver checking) https://gist.github.com/Redjumpman/3b49508ff7b86bbee678bfac1ddc2674#file-python-2-7-15-change-log-L264

PR #536, opened 2018-10-30
Add support for emulating physical devices to iotile-core

timburke:
## Overview

This PR is a first step that adds the concept of EmulatedDevice and EmulatedTile objects to `iotile-core`.  These form the basis emulating physical devices (and tiles).  The `virtual-device` script is augmented with additional command line parameters to support snapshoting the state of an emulated device as well as loading a saved snapshot and tracking changes to an emulated device's state during emulation.

## Expected Usage

The major expected use case of this feature is to spin up emulated devices for testing difficult-to-produce edge cases in physical device behavior.  The emulation is behavioral, so this will not catch issues in the physical devices themselves but rather in software components that interact with those physical devices.

This PR just lays the groundwork for the first useful feature which will be when `ReferenceController` and `ReferenceDevice` have enough emulated RPC functionality to do useful integration tests with.

### Scenario

A scenario is a preregistered handler function that is designed to prepare an EmulatedDevice or EmulatedTile into a specific internal state with arguments to allow the user to customize the precise details of the state.  It's expected that EmulatedDevices will define scenarios for most expected test cases that users may want to run on them.

### State

A `state` is a complete serialized snapshot of the internal state of an EmulatedDevice or EmulatedTile.  The exact state of the tile or device can be reconstructed by loading this serialized snapshot.

### EmulatedStateLog

For correctness verification and regression tracking purposes, EmulatedDevice objects can produce a log of all changes to their internal state that can be written out to a csv file and compared to future runs of the same test case.  EmulatedDevice and EmulatedTile objects can choose to register instance properties that form part of their internal state and writes to those properties will automatically be logged into the EmulatedStateLog.  For internal states that cannot be easily captures in a single property, emulated objects can also manually record a state change by calling `track_change`.

EmulatedDevices automatically incorporate the internal state logs of all of their constituent tiles (if they have any).

## Next Steps

- Move `ReferenceController` and `ReferenceDevice` to use `EmulatedTile` and `EmulatedDevice` and increase the RPCs implemented in `ReferenceController` until it is a useful emulation of a physical IOTile controller.
- Add appropriate EmulatedTile and EmulatedDevice classes along with their corresponding firmware implementations to provide for emulating production IOTile devices.


iotilecore/iotile/core/hw/virtual/emulation/__init__.py
```diff
@@ -0,0 +1,16 @@
+"""Base classes for virtual IOTile devices designed to emulate physical IOTile Devices.
+
+Classes that derive from EmulatedDevice and EmulatedTile have additional funtionality
+for saving and loading their state to provide for easy creation of testing scenarios.
+
+For example, you may have an IOTile Device that serves as a shock and vibration data
+logger.  In reality it would take time to load it up with real captured shocks and
+vibrations in order to test that we could properly create software that read those
+waveforms.  With an EmulatedDevice, you can just load in a `has 100 waveforms`
+scenario and use that as part of automated integration test cases.
```
dkarchmer: Can you define what scenarios and states are. I assume scenarios are just a collection of data, but I am less clear on what states are.

timburke:
> Very cool.
> 
> Curious, would a feature fit in the framework to, perhaps, reproduce a field configuration by "snapshotting" the state of a physical device and then loading that real snapshot in to the emulated device?

That could definitely be an interesting future feature.  I hadn't thought of it but it makes total sense.

iotilecore/iotile/core/hw/virtual/emulation/state_log.py
```diff
@@ -0,0 +1,102 @@
...
+
+        Args:
+            out_path (str): The path where we should save our current list of
+                changes.
+            header (bool): Whether we should include a header row in the csv
+                file.  Defaults to True.
+        """
+
+        # See https://stackoverflow.com/a/3348664/9739119 for why this is necessary
+        if sys.version < (3, 0, 0):
```
mattrunchey: alternatively we could just use the major from sys:
`if sys.version_info.major < 3`

timburke: Agreed, fixed now.

Issue #539, opened 2018-10-31
Update for compatibility with pyparsing 2.3.0


timburke:
It looks like pyparsing 2.3.0 is backwards incompatible even though it's marked as a minor update.  It came out 12 hours ago and it looks like it breaks `iotile-sensorgraph` and possibly also `iotile-build`.

Immediate fix is to pin the version specifier of pyparsing in both packages to `pyparsing~=2.2.0` but we should fix but we should also fix the incompatibility and move to using pyparsing 2.3.0.  

It looks from the release notes like they fixed a previous bug that we had already built a workaround to inside out code and the fix changes the behavior in a way that breaks the workaround.

PR #541, opened 2018-10-31
Split emulation functionality out into new iotile-emulate package

### LINKED ISSUE #538: Split IOTile emulation into separate package `iotile-emulate`
The scope of the emulation subsystem is going to get pretty large and it relies on behavior in both `iotile-core` and `iotile-sensorgraph` for proper controller emulation.  

I had been splitting the functionality between `iotile-core` and `iotile-sensorgraph` based on what dependencies it had but it makes the code harder to follow and is going to add a lot of bloat into `iotile-core` that is not necessary for normal use.

So, I propose creating an `iotile-emulate` package that depends on `iotile-core` and `iotile-sensorgraph` and moving all of the emulation code there.

timburke:
## PR Overview

This PR breaks all of the emulation related functionality out into a new `iotile-emulate` package.  Previously it was split between `iotile-core` and `iotile-sensorgraph` but this was leading to too much bloat in `iotile-core` and a difficult to follow logical separation between what parts were in each package.

Now everything is in `iotile-emulate` which depends on both `iotile-core` and `iotile-sensorgraph`.  See #538.

 ### Key Changes

- Creates `iotile-emulate` package
- Moves all emulation related code from `iotile-core` and `iotile-sensorgraph` into `iotile-emulate`.  
- Updates unit tests that referenced `reference_1_0` virtual device to work with it living in `iotile-emulate` now.
- Adds some additional helper classes to assist with state snapshotting of emulated devices.  `SerializableState` class is one example.
- Implements support for `config variables` into `EmulatedTile`
- Adds a base class for all non-controller emulated tiles `EmulatedPeripheralTile`.  
- Properly implements the correct tile lifecycle with reset, config variable streaming followed by start_app rpc from controller.  Adds a `DemoEmulatedDevice` for testing this process.
- Moves the ReferenceController to a mixin based architecture to allow for cleanly adding in all required controller subsystems.  Currently `RemoteBridge` is mostly implemented and `TileManager` is wireframed just enough to correctly support booting peripheral tiles.
- Adds a way to directly call RPCs on an EmulatedDevice class without needing to use a proxy and HardwareManager.  This is needed for internal use so that emulated tiles can talk to each other but also makes for simpler unit testing in some cases.

### Next Steps

> Next key milestone is having enough emulation in place to be able to mimic a complete Arch POD.

- This PR is basically just to get everything cleaned up so that work can start on `iotile-emulate` as a package.  The next big job is to start adding the relevant error code and well-known RPC constants into `iotile.emulate.constants` and remove any lingering hardcoded integer values from the code.
- I need to add remaining controller subsystems as mixins into ReferenceController
- I need to properly support sending the reset rpc to a tile.  Implementation is trivial if there is no background work happening on the tile but more complex tiles have a background thread that needs to be properly stopped and restarted after the reset.  

## Additional Fixes

- Pins `pyparsing` version in `iotile-build` and `iotile-sensorgraph` to avoid backwards incompatible change in pyparsing released yesterday.  See #539 

Closes #538


timburke:
@dkarchmer I updaetd and refactored the main README

PR #543, opened 2018-11-01
Add initial set of constant tables and increase test coverage.

timburke:
This PR just starts adding a list of global constants for well-known rpcs and errors codes that need to be implemented by emulated tiles.  

Attention was paid to properly documenting each RPC so that we can use the table also as the basis for future documentation of the semantics of each RPC.


iotilecore/iotile/core/hw/virtual/common_types.py
```diff
@@ -76,7 +76,7 @@ def pack_rpc_payload(arg_format, args):
         bytes: The packed argument buffer.
     """
 
-    code = _create_argcode(arg_format, args)
+    code = _create_respcode(arg_format, args)
```
timburke: Bugfix, this was incorrectly implemented before.

PR #549, opened 2018-11-03
Add test coverage of config_database and tile_manager subsystems.

### LINKED ISSUE #545: Support passing a DeviceAdapter class directly to HardwareManager
There are some times in unit testing where it would be easier to build a `DeviceAdapter` programmatically and then just pass it full built to `HardwareManager` to use instead of needing to serialize the configuration that you want to a json file or a base64 encoded string.

HardwareManager should support receiving a CMDStream or DeviceAdapter class on its `port` argument.  If it receives either it should just use it directly rather than building one from scratch assuming port is a string.

timburke:
Fixes a few bugs found during testing.  There is now test coverage on all RPCs in the config_database and tile_manager subsystems.

Additionally there is test covergae of config variable streaming to a peripheral tile and proper clearing of config variables on reset with streaming of new data from the controller.

@mattrunchey 
For ease of testing, this adds the ability to pass premade DeviceAdapter classes to HardwareManager.  This obviates in some cases the need to serialize device states and pass them as base64 encoded strings.

Closes #545

### DIFF
--- a/iotileemulate/test/test_reference.py
+++ b/iotileemulate/test/test_reference.py
@@ -2,9 +2,11 @@
 
 import pytest
 from iotile.core.hw import HardwareManager
+from iotile.core.hw.proxy.external_proxy import find_proxy_plugin
 from iotile.emulate.virtual import EmulatedPeripheralTile
 from iotile.emulate.reference import ReferenceDevice
 from iotile.emulate.constants import rpcs, Error
+from iotile.emulate.transport import EmulatedDeviceAdapter
 
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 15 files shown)


iotileemulate/test/test_reference.py
```diff
@@ -20,6 +22,25 @@ def reference():
...
+    """Get a reference device and connected HardwareManager."""
+
+    device = ReferenceDevice({})
+    peripheral = EmulatedPeripheralTile(11, b'abcdef', device)
+    peripheral.declare_config_variable("test 1", 0x8000, 'uint16_t')
+    peripheral.declare_config_variable('test 2', 0x8001, 'uint32_t[5]')
+
+    device.add_tile(11, peripheral)
+
+    adapter = EmulatedDeviceAdapter(None, devices=[device])
```
timburke: @mattrunchey For tests that you're writing in CoreTools, here's a way to avoid needing to serialize states into base64 and pass a giant string or create json files.  You can just directly create the virtual device classes and pass them into a precreated `VirtualDeviceAdapter` instance.

PR #556, opened 2018-11-05
Move RPC dispatch to a background workqueue

### LINKED ISSUE #554: Add ability to easily return error code from virtual RPC
Some emulated RPCs need to be able to return a non-0 error code from the RPC using its 1 byte status code.  There should be a simple exception they can raise to indicate the status code they want to send.

timburke:
In preparation for supporting EmulatedTiles that have their own background work going on, we need a general way of synchronizing RPCs to execute one at a time without requiring the RPC function to be called from a single thread.

This PR refactors the RPC dispatch logic to use a single background work-queue as the sole RPC executor.  All calls to `rpc`, `deferred_rpc` and `call_rpc` are not synchronized through this queue.  

Additional Fixes:
- Adds a lot more docstring descriptions of what is going on and why
- adds a `format_rpc` function that can print a nice string describing an RPC call-and-response including looking up a good name for the RPC (so `SET_CONFIG_VARIABLE` rather than just `RPC 0x2001`).  This is used for state tracking so that the produced changelog is easier to understand.
- Adds the background work queue thread to `iotile-core` for generic use along side the existing `StoppableWorkerThread`.  
- Adds the ability to directly set the return status from a virtual RPC.  This previously wasn't needed because normal Virtual devices don't use this mechanism to return errors but legacy RPCs do so it's needed for proper emulation.

Closes #554.

### DIFF
--- a/iotilecore/RELEASE.md
+++ b/iotilecore/RELEASE.md
@@ -10,6 +10,11 @@ All major changes in each released version of IOTileCore are listed here.
 - Add support for passing premade DeviceAdapter classes into HardwareManager
   (Issue #545).  This is very useful for setting up complex test scenarios.
 
+- Add support for WorkQueueThread to support dispatching work to background work
+  queues and inserting sequence points where all prior work needs to be
+  finished.
+
+
 ## 3.22.13
 
 - Add "watch_reports" command, which enables a user to live view reports from the connected device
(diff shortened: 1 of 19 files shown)


timburke:
@mattrunchey Good call on additional testing.  I added another set of test cases to cover ordering between sync and async calls to `dispatch`.

PR #574, opened 2018-11-08
Add initial sensor-graph subsystem

timburke:
This PR adds support for the `sensor-graph` subsystem to the reference controller.

This was pretty straightforward since `iotile-sensorgraph` already has a complete emulator for the underlying behavior.  This work just involved writing the RPC declarations for how you configure the emulator with the right error codes and moving everything to a background emulation thread.

**Additional Changes:**
  - Fixes the behavior of SensorGraph emulator to initialize constant streams with 0s.  This matches physical device behavior.
  - Adds a `wait_until_idle` to WorkQueueThread.  The docstring explains why this is useful in general.  In particular, it's necessary for properly unit testing `EmulatedDevice` since you need to know when the device has settled into a stable state after one of your inputs.  You can't just call `flush()` or `wait_deferred_rpcs()` since one of the rpcs added in the meantime might schedule another rpc.  So there needs to be a way to wait until the work queue is actually empty.
   - Adds the ability to enforce resource constraints on adding nodes and streamers to a `SensorGraph` according to the `DeviceModel` in use.
   - Adds a few more utility functions to `iotile-sensorgraph` objects to assist with using them from `iotile-emulate`.


iotilesensorgraph/iotile/sg/processors.py
```diff
@@ -35,9 +34,8 @@ def copy_all_a(input_a, *other_inputs, **kwargs):
 def copy_latest_a(input_a, *other_inputs, **kwargs):
     """Copy the latest reading from input a into the output.
 
-    All other inputs are skipped to that after this function
-    runs there are no readings left in any of the input walkers
-    even if no output is generated.
+    All other inputs are skipped to that after this function runs there are no
```
mattrunchey:
```suggestion
    All other inputs are skipped so that after this function runs there are no
```

Issue #587, opened 2018-11-12
Race condition in `iotile-emulate` unit test


timburke:
https://travis-ci.org/iotile/coretools/jobs/454095832#L594

There's a missing `device.wait_idle()` after reset in `test_config_database()`.

timburke:
@mattrunchey For tracking the issue.

PR #589, opened 2018-11-13
Fix string to byte encode error on py3

mattrunchey:


iotilecore/iotile/core/hw/virtual/virtualtile.py
```diff
@@ -38,7 +38,7 @@ def _check_convert_name(cls, name):
         if not isinstance(name, bytes):
             name = name.encode('utf-8')
         if len(name) < 6:
-            name += bytes(' ')*(6 - len(name))
+            name += bytes(' '.encode('utf-8'))*(6 - len(name))
```
timburke: Also I think this could be rewritten as simply `b' '*(6 - len(name))` as well

mattrunchey: Good point. Simplified.

mattrunchey:
I believe that test failure is an intermittent one due to the race condition issue #587 .

PR #594, opened 2018-11-27
Add parallel rpc flow for combining the 2-byte feature+command in to …

mattrunchey:
…a single command

rpc_v2 also enforces the use of arg_format and result_format, which changes the order for passing args through RPCs (python limitations mean we have to pass those arguments before *args and **kwargs, which are used in some rpc commands).

Right now it's only implemented for AdapterStream and CmdStream, but may need to add v2 methods to other streams if being used there as well.

---

I have converted the proxy calls in mqtt_forwarder and related rpc_dispatcher code to use the v2 format and it seems to function appropriately.


timburke:
## Slightly Larger Refactor

> While the initial PR does work, we can refactor things slightly to not need so much duplicated code while also preparing for getting rid of `rpc` over time and migrating everything to `rpc_v2`.

### Background Facts

> These statements of truth about what is and is not used anywhere else in the code base underlay the slightly different refactor I'm proposing.

1. The `RPCCommand` class is vestigial and not used anywhere except to pack integers into a bytes object using `struct.pack`.  In particular nothing else from that class is used and its `__str_` method is never used anywhere.  It is a very old holdover from before we supported bluetooth and RPCs were sent by serializing them into an ascii string and sending them over a serial port to an external I2C dongle that directly injected them on to the TileBus via a physical connection.

2. There is no direct use of the `CMDStream` methods anywhere except via `TileBusProxyObject` so we are free to change the `CMDStream` api without breaking anything (with one caveat noted below that is easy to update).

### Proposal

Given the above facts, we can do the following:

1. Change the `CMDStream.send_rpc` method to have the following signature:
    ```
     send_rpc(self, address, rpc_id, call_payload, **kwargs)
    ```
    
    With this change we can remove `RPCCommand` from `cmdstream.py` entirely.  This will mean that `TileBusProxyObject.rpc` is now in charge of packing arguments into a bytearray before passing it to `CMDStream.send_rpc`.

2. Fix the one other place that directly invokes `CMDStream.send_rpc` to follow the new argument format:
    https://github.com/iotile/coretools/blob/0d9b63ef27b96222450dd26f7714f6dfd52c74cf/iotilesensorgraph/iotile/sg/sim/hosted_executor.py#L55
  
3. Consolidate the `RPCCommand._format_args()` method into `TileBusProxyObject.rpc()` if there is no `arg_format` passed.  This is needed to ensure that it can call the new `CMDStream.send_rpc()` method with a binary bytes object rather than potentially multiple ints.  

4. Update `RecordedStream` that has the wrong signature for `_send_rpc`.  It's called with a single argument that's a bytes object, not multiple `*args`.

5.  Remove the entire `iotile.core.hw.commands` package that is no longer needed anywhere.  You'll need to remove a few references to it that aren't needed after the change in 1:
    https://github.com/iotile/coretools/search?q=RPCCommand&unscoped_q=RPCCommand

6. `RPCCommandV2` is no longer needed.

timburke:
@mattrunchey Let me know any questions you have on the proposal or anything that isn't clear.

mattrunchey:
> @mattrunchey Let me know any questions you have on the proposal or anything that isn't clear.

Sounds solid, I will work on it and push the changes to this PR

mattrunchey:
@timburke I *think* I took care of what was needed for this. Unsure that I've unraveled `RecordedStream` correctly, though. Pushed changes to the same commit just now.

iotilecore/iotile/core/hw/commands/rpc.py
```diff
@@ -1,98 +0,0 @@
-# This file is adapted from python code released by WellDone International
-# under the terms of the LGPLv3.  WellDone International's contact information is
-# info@welldone.org
-# http://welldone.org
-#
-# Modifications to this file from the original created at WellDone International
-# are copyright Arch Systems Inc.
-
-from builtins import str, int
```
timburke: These two imports need to be ported over to `TileBusProxyObject` now.  They are provided by the python `future` module to backport python 3 unicode features to python 2.  It's important because the check on line 101:

https://github.com/iotile/coretools/pull/594/files#diff-7a7c0562805a6437e11a53d9b89d8fc0R101

will be incorrect without it since on python 2 `str` and `bytes` are the same so it will try to `utf-8` encode binary data, which is why this test is failing:
https://travis-ci.org/iotile/coretools/jobs/460641356#L603

timburke: @mattrunchey

mattrunchey:
@timburke let me know if we're good to merge this

PR #608, opened 2018-12-14
Fix critical bug in binary node descriptor generation

timburke:
This PR fixes a critical trub generation bug in `iotile-sensorgraph`.  It also prepares for releasing a new `iotile-sensorgraph`  with the fix.

The bug resulted in an incorrect sgf being programmed into a POD when using a TRUB script instead of a snippet.  The bug only affected nodes that have a single input and had a non-trivial triggering condition on that input.  The result was that the triggering condition was ignored and the node would fire on all input changes.

See the diff below for the different TRUB output before and after the fix:

image

Closes https://github.com/iotile/production_devices/issues/101


timburke:
To prevent more issues like this in the future, this PR also changes the default behavior of the sensorgraph emulator to match the actual behavior of a device, which is that unused inputs default to `always` triggering instead of defaulting to `never` triggering.

PR #609, opened 2018-12-14
Add 'show_rpcs' option to iotile-updateinfo.

ekkizogloy:
This function just changes the priorities around so that the RPCs get displayed...


iotilecore/iotile/core/hw/update/script.py
```diff
@@ -113,8 +115,9 @@ def FromBinary(cls, script_data, allow_unknown=True):
 
             curr += total_length
```
timburke: I think it would be cleaner to just add a check here rather than fudging with the match quality because that is likely going to break a lot of other non-rpc records that use that behavior to find the appropriate record:

```python
try:
	if show_rpcs and record_type == SendRPCRecord.MatchType():
		record = SendRPCRecord.FromBinary(record_data, record_count)
    elif show_rpcs and record_type == SendErrorCheckingRPCRecord.MatchType():
       record = SendErrorCheckingRPCRecord.FromBinary(record_data, record_count)
	else:
    	record = UpdateRecord.FromBinary(record_data, record_count)
except DeferMatching as defer:
```

ekkizogloy:
@timburke @mattrunchey 
Changes made. Let me know what you think.

PR #616, opened 2019-01-10
Fix and extend UTCassigner handling of anchor points

timburke:
The UTCAssigner previously could only handle RTC based timestamps.  This adds built-in support for epoch based timestamps as well.  It also fixes the handling of automatic anchor points created from streams that are designated as containing anchor values.


iotilecore/iotile/core/hw/reports/utc_assigner.py
```diff
@@ -96,6 +98,19 @@ def convert_rtc(cls, timestamp):
         delta = datetime.timedelta(seconds=timestamp)
         return cls._Y2KReference + delta
 
+    @classmethod
+    def _convert_rtc_anchor(cls, reading):
+        """Conert a reading containin an RTC timestamp to datetime."""
```
mattrunchey:
```suggestion
        """Convert a reading containing an RTC timestamp to datetime."""
```

PR #617, opened 2019-01-11
Slightly cleanup iotile-emulate code and add support for asynchronous RPCs

timburke:
This PR merges some slight code cleanup of `iotile-emulate` that I had on a side branch and finishes initial support for asynchronous RPCs, which are RPCs whose handler function returns without giving a response.  From the caller's point of view, all RPCs (including these) are synchronous, so the caller remains blocked, however the actual response value is provided via a callback on `EmulatedDevice.finish_async_rpc` rather than the return value of the rpc handler itself.

This is an important feature for emulating many real-world tiles that extensively use asynchronous rpcs.

**Additional changes:**

- The reset handling logic of the emulator is slightly refactored to remove some race conditions
- The `EmulatedDeviceAdapter` is updated so that a device reset happens synchronously by the time a `reset` rpc returns.  This is important because it allows for race free unit tests that involve emulated device resets.  In physical devices there is a 1 second delay to give the device time to actually reset, which would be very slow in a unit testing context.  With this change, the delay is not needed for emulated devices and resets should now be race-free.
- Fixes to `WorkQueueThread` to properly allow for deferring the response to a work item.  This is the underlying feature used by asynchronous rpcs.
- Explicit checks for programming errors that would cause deadlocks are adding using a thread-local variable to determine if code is running on the rpc execution thread, which cannot block on an rpc or sychronously attempt to execute an rpc.

### DIFF
--- a/iotilecore/iotile/core/hw/reports/utc_assigner.py
+++ b/iotilecore/iotile/core/hw/reports/utc_assigner.py
@@ -120,6 +120,7 @@ class UTCAssigner(object):
         if uptime is None and utc is None:
             return
 
+        ##FIXME: if uptime is None, this will break
         if uptime & (1 << 31):
             if utc is not None:
                 return
--- a/iotilecore/iotile/core/hw/reports/utc_assigner.py
+++ b/iotilecore/iotile/core/hw/reports/utc_assigner.py
@@ -120,8 +120,7 @@ class UTCAssigner(object):
         if uptime is None and utc is None:
             return
 
-        ##FIXME: if uptime is None, this will break
-        if uptime & (1 << 31):
+        if uptime is not None and uptime & (1 << 31):
             if utc is not None:
                 return
 
--- a/iotileemulate/iotile/emulate/reference/reference_device.py
+++ b/iotileemulate/iotile/emulate/reference/reference_device.py
@@ -133,6 +133,7 @@ class ReferenceDevice(EmulatedDevice):
         consistent atomic restoration process.
 
         This method will block while the background restore happens.
+
         Args:
             state (dict): A previously dumped state produced by dump_state.
         """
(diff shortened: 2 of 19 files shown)


iotilecore/iotile/core/hw/reports/utc_assigner.py
```diff
@@ -120,6 +120,7 @@ def add_point(self, reading_id, uptime=None, utc=None, is_break=False):
         if uptime is None and utc is None:
             return
 
+        ##FIXME: if uptime is None, this will break
```
mattrunchey: just double checking -- is this just a note for the future or was it intended to be fixed here?

iotileemulate/iotile/emulate/reference/reference_device.py
```diff
@@ -77,6 +77,7 @@ def start(self, channel=None):
                 continue
 
             # Check and make sure that if the tile should start that it has started
+            # FIXME: Remove hardcoded run level constant
```
mattrunchey: same note on this

iotileemulate/iotile/emulate/virtual/emulated_device.py
```diff
@@ -46,31 +43,117 @@ def __init__(self, iotile_id, name):
...
 
     def _background_dispatch_rpc(self, action):
         """Background work queue handler to dispatch RPCs."""
 
         address, rpc_id, arg_payload = action
 
-        try:
-            exc_status = None
-            resp = None
+        # FIXME: Check for a queued RPC and return busy
```
mattrunchey: same

iotilecore/iotile/core/hw/reports/utc_assigner.py
timburke: Sorry that was just laziness.  I'm fixing it.

iotileemulate/iotile/emulate/reference/reference_device.py
timburke: Laziness again.  Updated.

iotileemulate/iotile/emulate/virtual/emulated_device.py
timburke: Converted to an issue since the feature is not trivial #619.

Issue #618, opened 2019-01-11
Add support for emulated device and tile discovery


timburke:
In order for emulation packages to be released along with firmware code, there needs to be way for them to found and installed for use with the emulator similar to how apps and proxies are found.

PR #623, opened 2019-01-14
Consolidate all extension importing code into a single place in iotile-core

### LINKED ISSUE #621: Refactor IOTile object and plugin discovery mechanism
As we look to support providing EmulatedTile objects and EmulatedDevice objects with firmware projects to include an easy-to-use emulator for all IOTile devices, we need to refactor how we declare and find iotile plugins.

This requires refactoring the `IOTile` object that is the basis for declaring what an IOTile project can contain and providing a general mechanism for finding plugins.

We should probably not list out all possible product types explicitly in the `IOTile` class either since it's not going to scale well.

timburke:
This PR cleans up the process by which `iotile-core` finds and imports plugins.  Over the years there have been many copies of the same importing code committed.  This PR removes them all from `iotile-core` and replaces them with 4 routines on `ComponentRegistry`:

- `load_extensions`: will find and load all extensions of a given type that are either pip installed or locally installed using `ComponentRegistry.add_component()`.  
- `load_extension`: will directly load and validate a single `.py` file that contains an extension.  This allows for directly specifying paths to extensions when required.  It also provides the basis for the automatic `load_extensions` method above.
- `register_extension`: will programmatically add an extension temporarily that will then be returned when is matches criteria specified in `load_extensions`.
- `clear_extensions`: clears any extensions previously registered using `register_extension`.

## Key Changes

- `ComponentRegistry` is now the single place for finding and loading extensions.  
- In order to speed up the creation of ComponentRegistry, all work related to reading its underlying key-value store is made lazy the first time it is actually needed.  This means that creating a ComponentRegistry is now very fast and the latency of reading the kv-store or importing `pkg_resources` is only incurred the first time you use a method that needs one of those two things.
- The `IOTile` object that contains parsed metadata for all locally installed IOTile components has been refactored to prepare for a world where there are more kinds of products that an IOTile component can provide.  In particular, all of the specific methods to list products of a certain kind have been combined into a single `find_products()` method.  The old methods are marked as deprecated and will be removed once the rest of coretools is updated not to need them and those versions get out into the wild for long enough.
- All places in `iotile-core` that directly imported python code have been refactored to use `load_extension` or `load_extensions` as needed. 
- Additional unit test coverage of the `IOTile` class to make sure that `find_products` works correctly in a wider array of circumstances.
- Previous ways to directly register an extension for unit testing purposes that were defined on `HardwareManager` have been generalized for any kind of extension and moved to `ComponentRegistry`.  Existing code has been ported over to the new interface.

Closes #621

Lays the ground work for #618

### DIFF
--- a/iotilecore/RELEASE.md
+++ b/iotilecore/RELEASE.md
@@ -4,6 +4,9 @@ All major changes in each released version of `iotile-core` are listed here.
 
 ## HEAD
 
+- Consolidate entry point related code into ComponentRegistry with a single
+  implementation of extension finding and importing.  Remove now redundant
+  reimplementations of the same code.
 - Fix support for STILL_PENDING flag in WorkQueueThread
 - Update AsynchronousRPCResponse exception to not need an `__init__` argument.
 
--- a/iotilecore/iotile/core/dev/__init__.py
+++ b/iotilecore/iotile/core/dev/__init__.py
@@ -1,2 +1,7 @@
-# This file is copyright Arch Systems, Inc.  
+# This file is copyright Arch Systems, Inc.
 # Except as otherwise provided in the relevant LICENSE file, all rights are reserved.
+
+from .iotileobj import IOTile
+from .registry import ComponentRegistry
+
+__all__ = ['IOTile', 'ComponentRegistry']
--- a/iotilecore/iotile/core/dev/iotileobj.py
+++ b/iotilecore/iotile/core/dev/iotileobj.py
@@ -3,6 +3,7 @@
 
 import itertools
 from collections import namedtuple
+from string import Template
 import json
 import os.path
 import sys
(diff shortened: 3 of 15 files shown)


timburke:
This refactor was important before we could make more progress on `iotile-emulate` because the actual emulated tiles need to be stored with their respective iotile components that they are emulating but we didn't support virtual tiles as products from an iotile component nor did we support finding them in `HardwareManager` if they were installed locally during development.

timburke:
@mattrunchey Thanks for finding the `find_proxy` references.  I searched for them to see where it was used but didn't find those for some reason.  We don't use those old tiles any more and that old usage should be updated anyway so I'm fine with those breaking.  It's a 1 line change if we pull them out of the archive and release them.

For `prepare_device.py` and related places in the rest of CoreTools that still need to be updated to use `ComponentRegistry.load_extensions` I wanted to wait a bit until the new iotile-core was released and then update the other packages and bump their `iotile-core` version requirement so they depend on the new functionality.  The tracking issue for this is here: https://github.com/iotile/coretools/issues/624

Issue #644, opened 2019-01-19
Consider migrating to importlib_resources for finding data file in packages


timburke:
As we move to consolidating all `entry_point` related code into `ComponentRegistry.load_extensions`, there are some places, most notably in `iotile-build` where we also explicitly load data files using `pkg_resources.resource_filename`.  This method does not have a direct, non-hacky analog that we could embed into `ComponentRegistry` and is just as slow as the rest of `pkg_resources`, potentially negating the speed benefits of #472, which the `load_extensions` refactor is designed to allow.

So, we need an alternative to `pkg_resources.resource_filename` and it looks like `importlib_resources` is the best alternative.  It is a backport of stdlib functionality introduced in Python 3.7 to 2.7 and 3.4+:
https://importlib-resources.readthedocs.io/en/latest/using.html

It has analog methods for most of what we need from `pkg_resources` in terms of file loading.

Issue #472, opened 2018-06-28
Add ability to freeze list of iotile plugins


timburke:
When running on an SBC with an SD card, enumerating the filesystem inside `pkg_resources` to find iotile plugins is very slow, adding 5-10 seconds to the start up time of any `iotile` command line script.  It's not possible to actually speed up `pkg_resources` but in situations where we know that the user is not installing / removing plugins (i.e. when the file system is read only), we can prescan all installed plugins and produce an `entry_points.txt` file similar to the component registry that contains all of the entry points.

If this file exists it can be consulted before `pkg_resources` is imported to speed up application load time:

See `pkg_resources` slow behavior, known bug, here:
https://github.com/pypa/setuptools/issues/926

timburke:
This will require changes across coretools:

- [ ] A new class in `iotile.core.utilities` called `EntryPointManager` that checks for `entry_points.txt` and loads it otherwise it loads `pkg_resources`.  It should have an `iter_entry_points` API as a drop in replacement for `pkg_resources` usage.
- [ ]  Port all Coretools packages to use EntryPointManager rather than directly using `pkg_resources`.

For this to be useful it has to be an all or nothing thing.  If anybody still imports `pkg_resources` the slow performance will return.

timburke:
It seems like a lot of larger projects have the same issue with `pkg_resources.iter_entry_points` and are settling on the alternative `entrypoints` package: https://github.com/anthony-tuininga/cx_Freeze/issues/216#issuecomment-435270807.

timburke:
This basic ability was added in PR #645.  You can now `iotile registry freeze` and `iotile registry unfreeze`, however there are a few remaining items:

1. This only affects parts of CoreTools that have been ported over to use the centralized extension system #624.  Any part of CoreTools still using `pkg_resources` directly will not pick up the frozen plugins.

2. There needs to be a clear logging message on program start saying that frozen plugins are being used so that users don't get confused and wonder why their newly pip installed packages aren't being detected.

PR #645, opened 2019-01-21
Improve support wheel building and speed up IOTile tool

timburke:
There are four main effects of this PR:

1. The process by which support wheels are built in `iotile-build` is extended to support more complicated packages that contain many files.  Previously, it was assumed that there would only be a few python modules in a support wheel that were just directly copied and bundled together as part of the build process.  However, this did not allow for those modules to import other modules reliably since the `load_extensions()` method of importing the modules during development would not create a proper package structure to allow for relative imports.  

    As support packages get more complicated because they include tile emulators, it's become important to allow for multiple modules that import each other so the python code in the support packages can be more easily maintained.  To this end, an IOTile component can now declare a `support_package` product that is a normal python package.  This package will become the support wheel and when individual entry points inside of it are loaded during development they will be loaded as proper submodules so that they can import each other as you would expect.  

    This makes complex support packages inside of a component much easier to design and test since they are just standard python packages.  Importantly, they are always imported using their fully unique support distribution name, even during development so that there are not collisions if multiple development components all put their support packages in the `python/support` directories (causing module shadowing on the `support` global package name).

2. It incrementally enhances `iotile-emulate` to allow for returning hardware versions and improves `SerializableState` to allow for serializing more complicated state trees without needing to specify explicit serializers and deserializers.  There is now support for serializing lists and dicts of objects automatically.

3. It removes all explicit references to pkg_resources from `iotile-core` and `iotile-build`, moving them both to using faster alternatives (the `entrypoints` package for entry points).  This is done primarily to allow for faster startup times in the iotile tool, which is quite slow on machines with slow disks such as those with SD cards.  See #644 and #472.  As part of this process the IOTile tool loading process was profiled and key culprits that are particularly slow to import (~ 100 ms each) are deferred.  This was `past.builtins` and `websockets`.  Neither was required to be loaded eagerly.  In testing, these changes were able to speed up the iotile tool's startup time by more than 2x.

4. It adds experimental support for speeding up the `iotile` tool in embedded contexts where the list of installed plugins is static.  In that scenario, there is no need to enumerate all installed packages looking for plugins since the list will not change.  You can instead just cache a list of plugins and look at that list instead.  Two new routines `iotile registry freeze` and `iotile registry unfreeze` are added to cache and remove the cache of plugins, respectively.  These methods are not yet ready for primetime use.  In particular, the rest of CoreTools needs to move completely off of `pkg_resources` before the cache will work completely.

### DIFF
--- a/iotilebuild/RELEASE.md
+++ b/iotilebuild/RELEASE.md
@@ -2,6 +2,11 @@
 
 All major changes in each released version of IOTileBuild are listed here.
 
+## HEAD
+
+- Refactor to remove all references to `pkg_resources`.  Uses centralized 
+  entrypoint loading system from iotile.core instead.
+
 ## 2.7.0
 
 - Add support for running pytest unit tests as part of the build process.
--- a/iotilebuild/RELEASE.md
+++ b/iotilebuild/RELEASE.md
@@ -4,6 +4,10 @@ All major changes in each released version of IOTileBuild are listed here.
 
 ## HEAD
 
+- Add support for complex python support wheels where there is an actual 
+  python package present in `python/support`.  Makes sure that the 
+  python package contents are properly copied into the output support wheel.
+
 - Refactor to remove all references to `pkg_resources`.  Uses centralized 
   entrypoint loading system from iotile.core instead.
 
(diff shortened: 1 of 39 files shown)


timburke:
Opened #647 to track the `past.builtins` issue.

PR #666, opened 2019-02-08
Feat contribution guide

mattrunchey:
I tagged a lot of people to review this because it has a larger goal.

The goal with this is to set an easily discoverable place for someone to learn, at a high level, how we develop software at Arch. It pulls from a few internal resources that we've used in part in the past, and will simplify and unify our approach.

At this point, the contribution guide is, to an extent, open to suggestions and changes. I want everyone to be on the same page on what we're proposing before setting it as a standard. 

Please remember that this PR is presented in an open forum, so certain conversations may need to be taken offline if they are closely tied to Arch internals. Please message me if you have any doubts about where your comment should go!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,127 @@
+# Contributing
+
+When contributing to this repository, please first discuss the change you wish to make via issue,
+email, or any other method with the owners of this repository before making a change. 
+
+Please note we have a code of conduct, please follow it in all your interactions with the project.
+
+## Coding Guidelines
+
+- We generally follow PEP8 for style, and you can see the `.pylintrc` file to check your code as you develop.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,127 @@
...
+```
+import logging
+
+logger = logging.getLogger(__name__)
+```
+
+## Development recommendation
+
+We use the [feature branch](https://docs.gitlab.com/ee/workflow/workflow.html) model with a rebase strategy. 
+Using this for development will make it easier for you tomerge pull requests.
```
lelithium: Small typo : `to merge`

dkarchmer:
Lets go ahead and merge this so the current guy trying to fix a bug can read it.

PR #667, opened 2019-02-10
Major iotile-emulate refactor to use asyncio

timburke:
## Overview

Previously we had the goal that `iotile-emulate` would work on both python 2 and python 3, which meant that it could not leverage any python 3 only features.  This led to a design using multiple threads for emulating each tile and a bunch of chained callbacks when we needed to defer activity that could not safely be performed in a given thread.  

The end result was that it was difficult to write the equivalent of a `main()` loop running on a tile which made it difficult to write an `EmulatedTile` that needed to emulate a tile with a non-trivial `main()` loop.

Given the pending EOL of python 2 and the significant maintainability advantage of using `asyncio` to implement the underlying emulator efficiently, we made the decision to move `iotile-emulate` to python 3 only (python 3.5+) and drop the goal of python 2 compatibility.

## Major Changes to iotile-emulate

> The key commits in this PR are squashed already to separate the large `iotile-emulate` change from the small tweaks to the other packages to avoid referencing iotile-emulate on python 2 and support releasing a python 3 only package.

- All emulation happens now in a single thread using the `EmulationLoop` class.  Previously there was an RPC dispatch thread, a separate clock thread and would have needed to be a thread for each tile with a non-trivial `main()` loop.  Now all of those threads are `coroutines` running inside of `EmulationLoop`.  The resulting code is significantly cleaner.
- All of the ReferenceController subsystems have been refactored to use coroutines for their background work rather than callbacks.
- The model for how an EmulatedPeripheralTile should implement a `main()` loop has been worked out and implemented in `DemoEmulatedTile` as a reference example with documentation of what you can do in an EmulatedTile and how to do it.
- The inheritance between `EmulatedTile` and `VirtualTile` has been broken since they no longer share any commonalities and `VirtualTile` assumes a threading based model of background activity.  
- The model of what it means for an emulated device to be `idle` has been adjusted to allow for background tasks.  This is a very important concept because it is what allows for a user to invoke an RPC in a test and then inspect the state of the emulated device without a race condition.  The `EmulationLoop` now has a concept of work queues that are tracked for when they are empty/have pending work and the EmulatedDevice is declared idle when there are no pending RPCs and no tracked work queue has unfinished work.

### Consequences of Dropping Python 2 in iotile-emulate

- There were a few unit tests in `iotile-ext-cloud`, `iotile-core` and `iotile-ship` that used an emulated device to perform integration tests.  Theses tests can only run now on python 3 and not python 2.  The first commit in this PR fixes all of those tests.
- The release scripts were not written to allow for a package to be released only on python 3, which means the `wheel` needs to be built on python 3.6, not python 2.7 since it is not a universal wheel.  The scripts were updated in the first commit of this PR as well.
- Firmware components that include an emulated tile will only be able to run unit tests against that emulator on python 3.
- Support wheels for tiles/devices that include emulators will only be able to use those emulators on python 3 CoreTools installations.

The net result of this move is that it marks a sea-change in CoreTools from being python 2.7 first with support for most things on python 3.6 to needing to be python 3 first with slowly ebbing support for python 2.7.  Given that python 2.7 is EOL in less than a year, this is probably an acceptable thing.  

From a user's standpoint, it means that installations on python 3.6 will be first class citizens now and any remaining code that is python 2.7 only will need to be updated.

From a development standpoint, moving additional packages to being python 3 only will give us a good opportunity to clean up a lot of legacy code whose structure has evolved gradually over the years but we now know what it should look like.

### DIFF
--- a/iotile_ext_cloud/test/conftest.py
+++ b/iotile_ext_cloud/test/conftest.py
@@ -1,5 +1,6 @@
 """Local fixtures for testing iotile-ext-cloud."""
 
+import sys
 import pytest
 from iotile.core.dev.registry import ComponentRegistry
 from iotile.cloud.cloud import IOTileCloud
@@ -102,6 +103,9 @@ def ota_cloud(mock_cloud_private_nossl):
 
 @pytest.fixture(scope="function")
 def simple_hw():
+
+    if sys.version_info < (3, 5):
+        pytest.skip('requires iotile-emulate on python 3.5+')
 
     simple_file = """{{
         "device":
--- a/iotileemulate/iotile/emulate/reference/reference_controller.py
+++ b/iotileemulate/iotile/emulate/reference/reference_controller.py
@@ -1,15 +1,14 @@
 """A reference implementation of the iotile controller that allows update verification."""
 
-from __future__ import print_function, absolute_import, unicode_literals
 import logging
-from future.utils import viewitems
+import asyncio
 from iotile.core.hw.virtual import tile_rpc, TileNotFoundError
 from iotile.core.hw.reports import IOTileReading
 from iotile.core.exceptions import ArgumentError
 from iotile.sg.model import DeviceModel
 from iotile.sg.parser import SensorGraphFileParser
 from iotile.sg.optimizer import SensorGraphOptimizer
-from ..virtual import EmulatedTile, synchronized
+from ..virtual import EmulatedTile
 from ..constants import rpcs, Error
 
 from .controller_features import (RawSensorLogMixin, RemoteBridgeMixin,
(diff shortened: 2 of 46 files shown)


iotileemulate/iotile/emulate/reference/reference_controller.py
```diff
@@ -2,14 +2,15 @@
 
 from __future__ import print_function, absolute_import, unicode_literals
```
mattrunchey: Do we still need these imports if we're only building against 3.5+?

iotileemulate/iotile/emulate/reference/reference_device.py
```diff
@@ -2,14 +2,12 @@
 
 import base64
 import logging
-import threading
-import time
+import asyncio
 from monotonic import monotonic
 from future.utils import viewitems
```
mattrunchey: another future we may be able to drop

iotileemulate/iotile/emulate/virtual/emulated_tile.py
```diff
@@ -5,12 +5,13 @@
 import struct
 import base64
 import logging
-from decorator import decorate
-from past.builtins import basestring
+import asyncio
+import inspect
 from future.utils import viewitems, viewvalues
```
mattrunchey: droppable futures

iotileemulate/iotile/emulate/reference/reference_controller.py
timburke: Good call.  I've remove all references to `future` and `__future__` from `iotile-emulate` and replaced the future routines with their python 3 equivalents.

iotileemulate/iotile/emulate/reference/reference_device.py
timburke: done

iotileemulate/iotile/emulate/virtual/emulated_tile.py
timburke: done.

iotile_ext_cloud/test/conftest.py
```diff
@@ -103,6 +104,9 @@ def ota_cloud(mock_cloud_private_nossl):
 @pytest.fixture(scope="function")
 def simple_hw():
 
+    if sys.version_info < (3, 5):
```
dkarchmer: Any advantage in supporting 3.5? In my experience, people (e.g. AWS) seem to have jumped from 3.4 all the way to 3.6. If we are moving to Python3, I think we should just go all the way to 3.6.

timburke: I'm onboard with running CoreTools on python 3.6 by default and encouraging users to always use the most recent stable interpreter but there don't seem to be any major changes between 3.5 and 3.6 that prevents us from getting 3.5 compatibility for free.  The only major advantage seems to be the presence of `f-strings` on python 3.6 to remove the need to call `str.format()` for simple string templating.  

When we go fully over to python 3 only there may be advantages in the better `typing` module for type annotations but other than that it seems like the major difference is some syntactic sugar for asynchronous list comprehensions:
https://docs.python.org/3/whatsnew/3.6.html

Supporting python 3.4 seemed like a burden because 3.5 adds `async/await` and is a huge step up in terms of writing coroutines but moving from 3.5 to 3.6 seems somewhat incremental.  Until we see a clear benefit to dropping python 3.5 entirely, I would rather default to supporting it.  Dropping support for python 3.4 was important to allow writing true coroutines with `async def`:

https://docs.python.org/3/whatsnew/3.6.html

mattrunchey: There are a few niceties of py3.6 with async/await that may just be nice to have, based on that what's new page, such as https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep525 (allowing await and yield in the same function body), as well as async generators. 

I agree that there aren't any game changers, but 3.6 gives us some good performance boosts on asyncio, because technically on 3.5 it's a provisional feature.


That being said, it's also less of a break to ask that users upgrade from 3.5 to 3.6, compared to asking for a move from 2.7 to a new major release, if we do change our minds later.

mattrunchey: Although, if we continue down that path, 3.7 *also* significantly enhances asyncio: 
https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-asyncio

Issue #652, opened 2019-01-26
Subtle issue with scanning... it is possible that there is a race condition.


ekkizogloy:
Traceback (most recent call last):
  File "scan_waterfall.py", line 80, in <module>
    sw.collect_data()
  File "scan_waterfall.py", line 74, in collect_data
    scan_results = hw.scan()
  File "<decorator-gen-28>", line 2, in scan
  File "/Users/leekkizogloy/arch/env/py27/lib/python2.7/site-packages/typedargs/utils.py", line 34, in _check_and_execute
    retval = func(*convargs, **convkw)
  File "/Users/leekkizogloy/arch/env/py27/lib/python2.7/site-packages/iotile/core/hw/hwmanager.py", line 891, in scan
    devices = self.stream.scan(wait=wait)
  File "/Users/leekkizogloy/arch/env/py27/lib/python2.7/site-packages/iotile/core/hw/transport/cmdstream.py", line 115, in scan
    return sorted(self._scan(wait=wait), key=lambda x: x['uuid'])
  File "/Users/leekkizogloy/arch/env/py27/lib/python2.7/site-packages/iotile/core/hw/transport/adapterstream.py", line 110, in _scan
    for name, value in viewitems(self._scanned_devices):
RuntimeError: dictionary changed size during iteration

mattrunchey:
I think this is inherent to the way we multithread. A quick fix would be to change that line to be `self._scanned_devices.copy()`.

timburke:
Yeah that does look like a bug.  I may be able to fix it on Sunday but a workaround for it would be to just try/catch the RuntimeError and ignore and try again.  It should happen somewhat infrequently I would guess but maybe not.

mattrunchey:
Actually, this part of the code only actually uses the values of the dictionary, so we can solve this in a 2/3 compatible way by converting the values to a list:

```
from future.utils import listitems
for value in listitems(self._scanned_devices):
  //logic
```

timburke:
There's a quick fix for this.  While it's not safe to iterate over the potentially changing `_scanned_devices` dict since it could change, we can just iterate over a copy:

https://blog.labix.org/2008/06/27/watch-out-for-listdictkeys-in-python-3

```python
for key, value in viewitems(self._scanned_devices.copy()):
    # Have your fun.
```

[...]

PR #670, opened 2019-02-11
Add basic event counting to the bled112 transport plugin

mattrunchey:
Counts total scan events, v1 parses, and v2 parses.

Usage:

If you have access to your adapter through the gateway, you can call these two functions, example:

`g.device_manager.adapters[0].get_scan_stats()`
`g.device_manager.adapters[0].reset_scan_stats()`

`get_scan_stats` now returns 5 things:
```
        return self._scan_event_count, self._v1_scan_count, self._v2_scan_count, self._device_scan_counts, 
(time_spent - self._last_reset_time)
```

And it will show you total scan events seen, as well as v1 specific adverts and v2. You should be able to use this information to deduct the rate of packets you're processing for a given environment.

It also tracks how many v1 and v2 advertising packets have been parsed since reset for each device.

Right now this "logging" is on by default. I can have a conditional that skips this logging unless we launch the adapter with debug level logging, as well.

### DIFF
--- a/transport_plugins/bled112/iotile_transport_bled112/bled112.py
+++ b/transport_plugins/bled112/iotile_transport_bled112/bled112.py
@@ -91,6 +91,13 @@ class BLED112Adapter(DeviceAdapter):
         self.connecting_count = 0
         self.maximum_connections = 0
 
+        self._scan_event_count = 0
+        self._v1_scan_count = 0
+        self._v1_scan_response_count = 0
+        self._v2_scan_count = 0
+        self._device_scan_counts = {}
+        self._last_reset_time = time.time()
+
         self._logger = logging.getLogger(__name__)
         self._logger.addHandler(logging.NullHandler())
 
@@ -123,6 +130,31 @@ class BLED112Adapter(DeviceAdapter):
                 found_devs.append(port.device)
 
         return found_devs
+
+    def get_scan_stats(self):
+        """Return the scan event statistics for this adapter
+
+        Returns:
+            int : total scan events
+            int : total v1 scan count
+            int : total v1 scan response count
+            int : total v2 scan count
+            dict : device-specific scan counts
+            float : seconds since last reset
+        """
+        time_spent = time.time()
+        return self._scan_event_count, self._v1_scan_count, self._v1_scan_response_count, \
+            self._v2_scan_count, self._device_scan_counts.copy(), \
+            (time_spent - self._last_reset_time)
+
+    def reset_scan_stats(self):
+        """Clears the scan event statistics and updates the last reset time"""
+        self._scan_event_count = 0
+        self._v1_scan_count = 0
+        self._v1_scan_response_count = 0
+        self._v2_scan_count = 0
+        self._device_scan_counts = {}
+        self._last_reset_time = time.time()
 
     def can_connect(self):
         """Check if this adapter can take another connection
@@ -537,6 +569,8 @@ class BLED112Adapter(DeviceAdapter):
         else:
             data = bytearray([])
 
+        self._scan_event_count += 1
+
         # If this is an advertisement packet, see if its an IOTile device
         # packet_type = 4 is scan_response, 0, 2 and 6 are advertisements
         if packet_type in (0, 2, 6):
@@ -544,12 +578,15 @@ class BLED112Adapter(DeviceAdapter):
                 return
 
             if data[22] == 0xFF and data[23] == 0xC0 and data[24] == 0x3:
+                self._v1_scan_count += 1
                 self._parse_v1_advertisement(rssi, string_address, data)
             elif data[3] == 27 and data[4] == 0x16 and data[5] == 0xdd and data[6] == 0xfd:
+                self._v2_scan_count += 1
                 self._parse_v2_advertisement(rssi, string_address, data)
             else:
                 pass # This just means the advertisement was from a non-IOTile device
         elif packet_type == 4:
+            self._v1_scan_response_count += 1
             self._parse_v1_scan_response(string_address, data)
 
     def _parse_v2_advertisement(self, rssi, sender, data):
@@ -581,6 +618,8 @@ class BLED112Adapter(DeviceAdapter):
         #   bit 5: Encryption key is user key
         #   bit 6: broadcast data is time synchronized to avoid leaking
         #   information about when it changes
+
+        self._device_scan_counts.setdefault(device_id, {'v1': 0, 'v2': 0})['v2'] += 1
 
         info = {'connection_string': sender,
                 'uuid': device_id,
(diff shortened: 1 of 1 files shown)


transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -124,6 +128,14 @@ def find_bled112_devices(cls):
 
         return found_devs
 
+    def get_scan_stats(self):
```
timburke: Per our new CoreTools contribution guide, these should have PEP8 docstrings :).

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -537,15 +551,19 @@ def _process_scan_event(self, response):
...
         # packet_type = 4 is scan_response, 0, 2 and 6 are advertisements
         if packet_type in (0, 2, 6):
             if len(data) != 31:
                 return
 
             if data[22] == 0xFF and data[23] == 0xC0 and data[24] == 0x3:
+                self._v1_scan_count += 1
                 self._parse_v1_advertisement(rssi, string_address, data)
             elif data[3] == 27 and data[4] == 0x16 and data[5] == 0xdd and data[6] == 0xfd:
+                self._v2_scan_count += 1
```
timburke: I would add a one more statistic here that will be good to keep track of:

1. On v1 packets there is both an advertisement and a scan_response (when active scanning).  Having a separate counter for v1 advertisements and v1 scan_responses will let us know how many times we get an advertisement but miss its scan response.  The second counter can go before the call to  `_parse_v1_scan_response`.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -124,6 +129,15 @@ def find_bled112_devices(cls):
...
         return found_devs
 
+    def get_scan_stats(self):
+        return self._scan_event_count, self._v1_scan_count, self._v2_scan_count
+
+    def reset_scan_stats(self):
+        self._scan_event_count = 0
+        self._v1_scan_count = 0
+        self._v2_scan_count = 0
+        self._device_scan_counts = {}
```
timburke: There is going to be a race here like #652 when you iterate over this dictionary since the dictionary size can change at any time from a background thread if a new device happens to be seen

mattrunchey: I should be a bit safer here, since I'm not doing a full iteration so much as a conditional. That is, the check for "does the key exist / add it, then +1 that key" should be sequentially uninterruptible by a call to reset_scan_stats(), since generally our GIL wouldn't likely interrupt.

However, I adjusted the implementation to use a single operation, which should be fully thread safe without needing a lock:
`self._device_scan_counts.setdefault(device_uuid, {'v1': 0, 'v2': 0})['v1'] += 1`

timburke: Understood.  I was more worried about if you were going to return `_device_scan_counts` to someone outside of this class.  That's what we did in #652 and can result in the caller having issues if they try to iterate over the result you returned them, e.g. to print it out and the result changes size during their iteration because a new device is seen.  

I think that would still be an issue even with the new code because the caller can take as long as they want to iterate over `self._device_scan_counts`.  A fix if the intent of `self._device_scan_counts()` is to hand it over to an outside caller would be to hand them a copy generated by `dict.copy()`:
https://blog.labix.org/2008/06/27/watch-out-for-listdictkeys-in-python-3

Actually this also resolves #652...

mattrunchey: If we had an iterator in the class that wanted to go over self._device_scan_counts, we should make sure that iterator does copy in its implementation as you mentioned above. The returned dictionary does not appear to be impacted on iteration though, because it does create a copy. I tested this by grabbing the returned dictionary, and resetting it during an iteration over the items:
```
>>> (s, v1, v1r, v2, devs, timedelta) = g.device_manager.adapters[0].get_scan_stats()
>>> for dev in devs.items():
...  print(dev)
...  g.device_manager.adapters[0].reset_scan_stats()
... 
(2779, {'v1': 0, 'v2': 170})
(2783, {'v1': 583, 'v2': 0})
(2092, {'v1': 89, 'v2': 0})
(2780, {'v1': 0, 'v2': 167})
(442, {'v1': 196, 'v2': 0})
(263, {'v1': 75, 'v2': 0})
(1751, {'v1': 85, 'v2': 0})
(2778, {'v1': 0, 'v2': 165})
(1330, {'v1': 93, 'v2': 0})
(1877, {'v1': 12, 'v2': 0})
(2093, {'v1': 46, 'v2': 0})
(1443, {'v1': 38, 'v2': 0})
(1473, {'v1': 36, 'v2': 0})
(2215, {'v1': 79, 'v2': 0})
(1339, {'v1': 0, 'v2': 92})
(557, {'v1': 46, 'v2': 0})
(2095, {'v1': 100, 'v2': 0})
(2633, {'v1': 0, 'v2': 39})
(2098, {'v1': 10, 'v2': 0})
(414, {'v1': 6, 'v2': 0})
(1337, {'v1': 1, 'v2': 0})
(1875, {'v1': 1, 'v2': 0})
```

timburke:
Try instead modifying the dictionary by adding a new entry. You should see
it fail.

mattrunchey:
Yea I ran that case right after commenting, my mistake. I was hung up on the fact that `copy` wouldn't truly create a new object if some of the items in the dictionary were lists that you then modified by reference, i.e. 

```
d2 = d1.copy()

# d1['somekey'][0] = 'oldval'

d1['somekey'][0] = 'newval'

# d2['somekey][0] = 'newval'
```

So I just have tokeep that in mind when using the contents of a copied dict.

slack0:
> Yea I ran that case right after commenting, my mistake. I was hung up on the fact that `copy` wouldn't truly create a new object if some of the items in the dictionary were lists that you then modified by reference, i.e.
> 
> ```
> d2 = d1.copy()
> 
> # d1['somekey'][0] -> 'oldval'
> 
> d1['somekey'][0] = 'newval'
> 
> # d2['somekey][0] -> 'newval'
> ```
> So I just have tokeep that in mind when using the contents of a copied dict.

Is `copy.deepcopy(d1)` the right thing to do here?

mattrunchey:
> > Yea I ran that case right after commenting, my mistake. I was hung up on the fact that `copy` wouldn't truly create a new object if some of the items in the dictionary were lists that you then modified by reference, i.e.
> > ```
> > d2 = d1.copy()
> > 
> > # d1['somekey'][0] -> 'oldval'
> > 
> > d1['somekey'][0] = 'newval'
> > 
> > # d2['somekey][0] -> 'newval'
> > ```
> > So I just have tokeep that in mind when using the contents of a copied dict.
> 
> Is `copy.deepcopy(d1)` the right thing to do here?

I think that since all of our values are integers, a `copy` has the desired effect as `int` is immutable and will create a new object to contain it (that is, updating the original dict's int value won't have an effect on the copied dict's int value).

If we wanted to have the dict values as mutable objects such as lists, we would want to do a deepcopy or use something immutable like putting it in to a tuple, or doing a full deep copy.

PR #678, opened 2019-02-13
Add rpc_v2 arg packing validation

### LINKED ISSUE #676: rpc_v2 does not appear to validate its arguments
I accidentally configured an RPC in a way that should fail but it didn't, it silently sent just a partial result.

I had code like the following:

```
error, = self.rpc_v2(0x8091, "BB12s", "L", index, int(force), result.encode())
```

However, `result.encode()` produced a 15-byte encoded value.  There should have been an exception saying that the arguments could not be encoded properly.  There was no exception and it's not clear what was sent but I assume it was the first 12-bytes of the 15-byte value.

Similarly, I had a previous example where I forgot to append an argument and the RPC still worked but sent nothing:

```
self.rpc_v2(0xAAAA, "L", "")
```

This should fail because it takes 1 argument for the "L" and not are provided but it appears to succeed silently.  In case it matters this was on python 3 but I don't think that it is the issue.

mattrunchey:
Fix rpc_v2 not checking that args were present before sending the rpc

Add some rpc tests

Fixes #676

### DIFF
--- a/iotilecore/iotile/core/hw/proxy/proxy.py
+++ b/iotilecore/iotile/core/hw/proxy/proxy.py
@@ -148,8 +148,11 @@ class TileBusProxyObject(object):
         """
         if args:
             packed_args = pack_rpc_payload(arg_format, list(args))
-        else:
+        elif arg_format == "":
             packed_args = b''
+        else:
+            raise IOTileException("Arg format expects arguments to be present", arg_format=arg_format, args=args)
+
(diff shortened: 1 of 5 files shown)


mattrunchey:
I can't tell why the tests I've added are failing, I don't seem to be registering the proxy object correctly.

mattrunchey:
I'm guessing it's because I didn't change the proxy name (overlapped with another `test01`). Changing the signature to `test42` seems to bypass it.

PR #681, opened 2019-02-13
Fix bug with sharing extensions between ComponentRegistry objects

vvynohradov:
I found that in clear_extensions() function, we didn't clear _registered_extensions object, we are assigning a new one to it, so when we are creating 2 ComponentRegistry objects, and adding an extension to the each one, second one would have an extension that we added to the first one.

### DIFF
--- a/iotilecore/RELEASE.md
+++ b/iotilecore/RELEASE.md
@@ -7,6 +7,7 @@ All major changes in each released version of `iotile-core` are listed here.
 - Add support for `emulated_tile` product to be included in an IOTile Component.
   This is necessary now that `iotile-emulate` no longer supported python 2 and
   requires asyncio inside its emulated tiles.
+- Fix bug with sharing extensions between ComponentRegistry objects
 
 ## 3.26.5
 
--- a/iotilecore/iotile/core/dev/registry.py
+++ b/iotilecore/iotile/core/dev/registry.py
@@ -209,7 +209,7 @@ class ComponentRegistry(object):
         """Clear all previously registered extensions."""
 
         if group is None:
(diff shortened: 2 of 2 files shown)


mattrunchey:
I'm not certain this is the behavior intended by the script. I need to look a bit more at the extension code to see what's going on here.

mattrunchey:
Ok, I think i finally wrapped my head around why this particular spot is breaking and not in other instances where we invoke self._registered_extensions instead of Class._registered_extensions. It could probably have been implemented a bit cleaner in the first place, that is, our mixing of class and instance variables would have helped, because it's not obvious at the start that this would create a new instance rather than using the class's identically named variable.

As a simpler example, just for posterity, demonstrating this unwanted behavior:
```
class LoadThing():
    a = {}
    b = {}

    def __init__(self):
        pass

    def clear_a(self):
        self.a = {}

    def clear_b(self):
        LoadThing.b = {}
```

The following behavior that we don't want is present:
```
>>> l1 = l()
>>> l2 = l()
>>> l1.a[1] = 2
>>> l1.a
{1: 2}
>>> l2.a
{1: 2}
>>> l2.clear_a()
>>> l2.a
{}
>>> l1.a
{1: 2}

>>> l1.b[1] = 2
>>> l1.b
{1: 2}
>>> l2.b
{1: 2}
>>> l2.clear_b()
>>> l2.b
{}
>>> l1.b
{}
```


I made sure the rest of registry.py wasn't creating a new self._registered_extensions object as well, and we seem to always call it with a subkey rather than instantiating a whole entire new object, so that's good.

Issue #626, opened 2019-01-15
Better error message when an entry_point is incorrect


timburke:
```
  File "/Users/mrunchey/gitrepos/coretools/iotilecore/iotile/core/scripts/iotile_script.py", line 208, in main
    finished = shell.invoke(line)
  File "/Users/mrunchey/gitrepos/typedargs/typedargs/shell.py", line 493, in invoke
    val, line, finished = self.invoke_one(line)
  File "/Users/mrunchey/gitrepos/typedargs/typedargs/shell.py", line 456, in invoke_one
    val = func(*posargs, **kwargs)
  File "<decorator-gen-18>", line 2, in __init__
  File "/Users/mrunchey/gitrepos/typedargs/typedargs/utils.py", line 34, in _check_and_execute
    retval = func(*convargs, **convkw)
  File "/Users/mrunchey/gitrepos/coretools/iotilecore/iotile/core/hw/hwmanager.py", line 85, in __init__
    self.stream = self._create_stream(adapter)
  File "/Users/mrunchey/gitrepos/coretools/iotilecore/iotile/core/hw/hwmanager.py", line 814, in _create_stream
    return AdapterCMDStream(adapter_factory(port), port, conn_string, record=self._record)
  File "/Users/mrunchey/gitrepos/coretools/iotilecore/iotile/core/hw/transport/virtualadapter.py", line 137, in __init__
    loaded_dev = self._load_device(name, config)
  File "/Users/mrunchey/gitrepos/coretools/iotilecore/iotile/core/hw/transport/virtualadapter.py", line 205, in _load_device
    for device_name, device_factory in reg.load_extensions('iotile.virtual_device', class_filter=VirtualIOTileDevice):
  File "/Users/mrunchey/gitrepos/coretools/iotilecore/iotile/core/dev/registry.py", line 158, in load_extensions
    ext = entry.load()
  File "/Users/mrunchey/ota_dev/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2332, in load
    return self.resolve()
  File "/Users/mrunchey/ota_dev/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2338, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
ModuleNotFoundError: No module named 'arch_gateway.config_device'
```

`load_extensions` should have a `try/catch` that prints more information about which entry point or distribution was incorrect.

timburke:
[...]

PR #682, opened 2019-02-13
Add exception handling in load_extensions(#626)

DimaWittmann:
Change load_extensions so that when an exception on loading a module is
raised it got caught and warning is logged in order to avoid breaking
everything if there is an error importing an extension.


iotilecore/RELEASE.md
```diff
@@ -8,6 +8,12 @@ All major changes in each released version of `iotile-core` are listed here.
   This is necessary now that `iotile-emulate` no longer supported python 2 and
   requires asyncio inside its emulated tiles.
 
+## 3.26.6
```
mattrunchey: If you want to push a new release tag, you need to move all of the existing stuff in ##HEAD to that block also, since those commits will be included in the new version of the software.

mattrunchey: So my recommendation is that it's probably better move your changes back to HEAD for now, since we do have pending HEAD changes already that we might want to review before releasing a version.

DimaWittmann: Thank you very much for your recommendations, unfortunately I was out of reach to my laptop and did not fulfill them. 
 Thank you one again for help.

Issue #657, opened 2019-01-30
ctrl+C breaks in `iotile-gateway` (and possibly others) if you accidentally double book the bled112 dongle


mattrunchey:
I was running `iotile-gateway` configured with bled112 as an adapter from one terminal.

In another terminal, I attempted to invoke `virtual_device bled112 virtual_device.py` . Since it, by default, tries to use the first bled112 dongle on the system (I have two and can avoid this by specifying the usbdev). A HardwareError ultimately gets raised, but the unexpected part is this seems to render exiting cleanly impossible (I had to ctrl+x and then kill -9 the pid). It would be nice if we could not have the ctrl+c handler break.

```
Traceback (most recent call last):
  File "/Users/mrunchey/gitrepos/coretools/transport_plugins/bled112/iotile_transport_bled112/async_packet.py", line 75, in ReaderThread
    chunk = bytearray(filelike.read(remaining_length - len(remaining)))
  File "/Users/mrunchey/ap_dev/lib/python3.6/site-packages/serial/serialposix.py", line 509, in read
    raise SerialException('read failed: {}'.format(e))
serial.serialutil.SerialException: read failed: device reports readiness to read but returned no data (device disconnected or multiple access on port?)
^CC-2019-01-30 11:25:33,702-main      :23   In quit signal handler.
C-2019-01-30 11:25:33,702-main      :26   Calling stop on gateway loop
C-2019-01-30 11:25:33,703-gateway   :165  Stopping gateway
I-2019-01-30 11:25:33,703-gateway   :166  Stopping gateway agents
C-2019-01-30 11:25:33,703-gateway   :174  Stopping device adapters
E-2019-01-30 11:25:33,704-bled112_cmd:156  Error stopping scan for devices, error=2
E-2019-01-30 11:25:33,704-gateway   :179  Error stopping device adapters
Traceback (most recent call last):
  File "/Users/mrunchey/ap_dev/lib/python3.6/site-packages/iotilegateway/gateway.py", line 177, in _stop_loop
    self.device_manager.stop()
  File "/Users/mrunchey/ap_dev/lib/python3.6/site-packages/iotilegateway/device.py", line 71, in stop
    adapter.stop_sync()
  File "/Users/mrunchey/gitrepos/coretools/transport_plugins/bled112/iotile_transport_bled112/bled112.py", line 140, in stop_sync
    self.stop_scan()
[...]

PR #683, opened 2019-02-13
657 issue fix

vvynohradov:
Added exclusive variable in the **Serial** constructor(Pyserial module). This both works on windows and on linux. On windows, it just ignores this flag, while on linux, it creates a lock file.

### DIFF
--- a/transport_plugins/bled112/RELEASE.md
+++ b/transport_plugins/bled112/RELEASE.md
@@ -2,6 +2,10 @@
 
 All major changes in each released version of the bled112 transport plugin are listed here.
 
+## HEAD
+
+- Ctrl+C breaks if you accidentally double book the bled112 dongle fix(#657)
+
 ## 1.8.0
 
 - Update virtual interface for compatibility with new iotile-core version that
--- a/transport_plugins/bled112/iotile_transport_bled112/bled112.py
+++ b/transport_plugins/bled112/iotile_transport_bled112/bled112.py
@@ -100,8 +100,7 @@ class BLED112Adapter(DeviceAdapter):
 
         self._logger = logging.getLogger(__name__)
         self._logger.addHandler(logging.NullHandler())
-
-        self._serial_port = serial.Serial(port, 256000, timeout=0.01, rtscts=True)
+        self._serial_port = serial.Serial(port, 256000, timeout=0.01, rtscts=True, exclusive=True)
         self._stream = AsyncPacketBuffer(self._serial_port, header_length=4, length_function=packet_length)
         self._commands = Queue()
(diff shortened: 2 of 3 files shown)


mattrunchey:
The lock also seems to have possibly long-term effects on the socket. I have been locked out of my port for a long time due to a program crash:
```
Traceback (most recent call last):
  File "/Users/mrunchey/gitrepos/coretools/iotilegateway/iotilegateway/gateway.py", line 123, in run
    adapter = adapter_class(port_string)
  File "/Users/mrunchey/gitrepos/coretools/transport_plugins/bled112/iotile_transport_bled112/bled112.py", line 99, in __init__
    self._serial_port = serial.Serial(port, 256000, timeout=0.01, rtscts=True, exclusive=True)
  File "/Users/mrunchey/newdev/lib/python3.7/site-packages/serial/serialutil.py", line 240, in __init__
    self.open()
  File "/Users/mrunchey/newdev/lib/python3.7/site-packages/serial/serialposix.py", line 272, in open
    self._reconfigure_port(force_update=True)
  File "/Users/mrunchey/newdev/lib/python3.7/site-packages/serial/serialposix.py", line 312, in _reconfigure_port
    raise SerialException(msg.errno, "Could not exclusively lock port {}: {}".format(self._port, msg))
serial.serialutil.SerialException: [Errno 35] Could not exclusively lock port /dev/cu.usbmodem11: [Errno 35] Resource temporarily unavailable

```

mattrunchey:
Found it, you should also change the other reference that attempts to load its serial device without the lock here as well:
https://github.com/iotile/coretools/blob/a5f2a846c72cc5995994ce78d94f20a2f80e9198/transport_plugins/bled112/iotile_transport_bled112/virtual_bled112.py#L80

transport_plugins/bled112/RELEASE.md
```diff
@@ -2,6 +2,10 @@
 
 All major changes in each released version of the bled112 transport plugin are listed here.
 
+## 1.8.1
```
mattrunchey:
```suggestion
## HEAD
```

vvynohradov:
@mattrunchey , thank you for your feeback and help, I have updated my branch. I will also research ctrl+c  problem. Where do you think it can appear again?

mattrunchey:
@vvynohradov I'm not sure right now which path we'd take to get in to a bad state, but it'd more likely be something that might corrupt the datastream on the dongle itself. We should move the discussion to #657 . It's more about not having a clear way to tell, in "stop_scan", how to shut down in one case but stay open in another case even when the same behavior is present right now.

PR #686, opened 2019-02-15
Created Lock() for race in scan(#652)

vvynohradov:
Created Lock() in scan() function


mattrunchey:
You should use the `with` contextmanager when possible instead of try/finally for locks.

For now let's put this on hold, though. We need to analyze if a lock is more efficient than a workaround like copy.

vvynohradov:
Ok, sure

PR #692, opened 2019-02-26
Bump iotile-core major version to drop python2 support. 

mattrunchey:
Submitting this separately from the other packages due to their dependency on it to test, and just for clarity. If there's another way to bundle everything together, happy to amend this PR to include those.

## Overview

We are EOL for python2 in preparation for the 2020 upstream EOL. Moving future dev work to python3 only helps as a forcing function to fix any remaining incompatibilities and lets us take advantage of py3 only features without breaking workflows when you support both.

## Major Changes

- iotilecore package dependency updates (some of them were quite out of date and likely invalid)
- Adjust package release to be non-universal py3 only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,13 @@ cache:
   directories:
   - ~/arm_gcc
   - ~/arm_qemu
-sudo: false
-python:
-- '2.7'
-- '3.6'
+matrix:
+  include:
+    - python: '3.5'
+    - python: '3.6'
+    - python: '3.7'
+      dist: xenial
+      sudo: true
 before_install:
 - pwd
 - cd ~
@@ -26,13 +29,6 @@ script:
 - tox -r
 
 deploy:
-  - skip_cleanup: true
-    provider: script
-    script: python scripts/release.py $TRAVIS_TAG
-    on:
-      branch: master
-      tags: true
-      condition: $TRAVIS_PYTHON_VERSION = "2.7"
   - skip_cleanup: true
     provider: script
     script: python scripts/release.py $TRAVIS_TAG
--- a/README.md
+++ b/README.md
@@ -10,8 +10,8 @@
 - [Installing Support for IOTile Based Devices](#installing-support-for-iotile-based-devices)
 - [Continuous Deployment](#continuous-deployment)
 - [Manually Releasing](#manually-releasing)
-- [Contributing](#Contributing)
-- [Code of Conduct](#Code of Conduct)
+- [Contributing](#contributing)
+- [Code of Conduct](#code-of-conduct)
 - [License](#license)
 
 <!-- /MarkdownTOC -->
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,9 @@ environment:
     # The list here is complete (excluding Python 2.6, which
     # isn't covered by this document) at the time of writing.
 
-    - PYTHON: "C:\\Python27"
+    - PYTHON: "C:\\Python35"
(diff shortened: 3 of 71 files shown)


mattrunchey:
@timburke Shouldn't the deploy check in travis.yml take care of not double-deploying for both tests?
```
deploy:
  - skip_cleanup: true
    provider: script
    script: python scripts/release.py $TRAVIS_TAG
    on:
      branch: master
      tags: true
      condition: $TRAVIS_PYTHON_VERSION = "3.5"
```

timburke:
> @timburke Shouldn't the deploy check in travis.yml take care of not double-deploying for both tests if we specify 3.5 and 3.6 as test targets?
> 
> ```
> deploy:
>   - skip_cleanup: true
>     provider: script
>     script: python scripts/release.py $TRAVIS_TAG
>     on:
>       branch: master
>       tags: true
>       condition: $TRAVIS_PYTHON_VERSION = "3.6"
> ```

@mattrunchey Yep you're correct there's no issue if that is the condition tag.

mattrunchey:
@timburke 

Ok. I believe that I pushed all packages up a major version (except iotile-emulate, I moved that to 0.4.0), and travis+appveyor are running 3.5/6/7 builds now. 72 files changed!

Well, appveyor should pass, but I built up a giant queue of tests and some of the old ones are taking 30 minutes due to the gateway hang on 3.5, so that might take a bit of time to resolve.

In theory, the `travis.yml` should only release one package with the condition on TRAVIS_PYTHON=3.6, but we might want to double check before merging. 

There were a few minor bugfixes that I put in to get the builds to pass, but I didn't detail them all in the release notes (they were things like changing `rb` to `r`, and fixing a tox dependency that was impacting the availability of some of the proxy objects).

.travis.yml
```diff
@@ -26,13 +29,6 @@ script:
...
-  - skip_cleanup: true
-    provider: script
-    script: python scripts/release.py $TRAVIS_TAG
-    on:
-      branch: master
-      tags: true
-      condition: $TRAVIS_PYTHON_VERSION = "2.7"
   - skip_cleanup: true
     provider: script
     script: python scripts/release.py $TRAVIS_TAG
```
timburke: Do you not still need the `tags: true` line here?  That was supposed to only deploy when we tagged a release rather than every commit to master

scripts/release.py
```diff
@@ -16,7 +16,7 @@
 import requests
 import setuptools.sandbox
 from twine.commands.upload import upload
-from components import comp_names
+from .components import comp_names
```
timburke: Are you sure this works?  I would have expected this to fail with a 'relative import attempted outside of a package' but maybe that's a python 2 only error.  You can run the scripts manually without releasing anything to test this by doing, for example:

```
python scripts/release.py --check iotilecore-4.0.0
```

.travis.yml
mattrunchey: I double checked and "tags: true" is still present for the 3.6 release, the 2.7 one was just completely gutted.

scripts/release.py
mattrunchey: Oh, good catch. I assumed it was a package without more than glancing at it (and my IDE complained that it was invalid syntax, which seems like I need to figure out w.r.t. my pycharm config)

.travis.yml
timburke: Understood, got it, I missed that part.

mattrunchey:
Closes #693 
Closes #694 
Closes #695

PR #701, opened 2019-02-27
Improve support for emulated tiles

timburke:
## Overview

This PR makes patch updates to `iotile-emulate`, `iotile-core` and `iotile-build`.  In core and build it fixes support for `support_package` products and extends `iotile build` to be able to add entry points into support wheels for `emulated_tile` products.

In `iotile-emulate` it fixes a bug where the streaming subsystem of the reference controller implementation never released a streamer after it finished streaming a report, causing it to not be able to stream more than one report per streamer.  Test coverage has been added to verify the fix.


iotileemulate/iotile/emulate/virtual/peripheral_tile.py
```diff
@@ -62,7 +62,11 @@ def __init__(self, address, device):
 
         self._logger.info("Tile %s at address %d is starting from reset", self.name, self.address)
 
-        address, run_level, debug = await self._device.emulator.await_rpc(8, rpcs.REGISTER_TILE, *self._registration_tuple())
+        try:
+            address, run_level, debug = await self._device.emulator.await_rpc(8, rpcs.REGISTER_TILE, *self._registration_tuple())
+        except:
```
mattrunchey: This bare except might also want a pylint:disable

timburke: I just checked this because I was wondering why pylint wasn't flagging it.  `pylint` doesn't care about `bare-except` if you `raise` the exception again since you're not chomping it so this is fine without a `pylint:disable`

mattrunchey: Oh, good to know!

Issue #443, opened 2018-04-13
qemu-system-gnuarmeclipse is broken on mac os x


timburke:
There is a bug internal in the program that has a hardcoded path for where to look for device SVD files.  A workaround is to specify `-L <path>` on the command line only on mac os x so that the device files are found correctly.

timburke:
Investigating this in the QEMU source code shows:

`qemu` tries to find the SVD json files by calling `qemu_find_file`:
https://github.com/gnu-mcu-eclipse/qemu/blob/de2bf74b160025c9744c0506ac2c96a7040afecb/hw/cortexm/mcu.c#L111
```c
        const char *svd_full_name = qemu_find_file(QEMU_FILE_TYPE_DEVICES,
                capabilities->svd_file_name);
        if (svd_full_name == NULL) {
            error_printf("JSON SVD file '%s' not found.\n", svd_full_name);
            exit(1);
        }
```

- `qemu_find_file` is implemented here:
https://github.com/gnu-mcu-eclipse/qemu/blob/de2bf74b160025c9744c0506ac2c96a7040afecb/vl.c#L2532

It has a list of `data_dir` folders to search through and for each one is appends `devices/` to the path since it is called with `QEMU_FILE_TYPE_DEVICES`.  You can add an entry to `data_dir` by passing `-L <data dir>` on the command line.  However, this `<data dir>` has to be the **parent folder** containing a devices subfolder since the `devices/` extension is automatically appended to whatever you enter by `qemu_find_file`

PR #704, opened 2019-02-28
Move windows and mac builds from appveyor to azure pipelines

mattrunchey:
Appveyor has been terribly slow as of late, and Azure Pipelines offers a 10-concurrent-build setup for open source projects that can run on hosted Windows and OSX. 

The goal of this is to partially make a move for validating behavior on windows and mac, but keeping the linux build on travis (and continuing to do releases from travis for the time being).

Pipeline actually passes in azure for both mac and windows. If this change gets approved I'll also adjust the required checks for this repo to require passing the Azure builds instead of the appveyor ones. We will still deploy from travis for now on master tags.


mattrunchey:
Apparently #443 is fixed now. At least, on the Azure instances, you can tap the new formula and `iotilebuild` works:
```

  - script: |
      brew tap PX4/px4
      brew install px4/px4/gcc-arm-none-eabi
      which arm-none-eabi-gcc
```

mattrunchey:
> Looks good. Is there an `azure` badge we can add to the README in place of the appveyor one?

Probably, I'll check how to and add that. Need to fix the travis build, the change I made to the tox.ini seems to have bypassed the native-ble install on linux because it's no longer identifying the env with `linux_only`.

mattrunchey:
On another note, we should probably have smarter test triggers so that every package doesn't retest when we change unrelated code. We shouldn't need to test `iotile_ext_cloud` when we change `iotile-emulate`, for example, but I'm not quite sure how we'd accomplish this.

PR #712, opened 2019-03-01
Completely remove past.builtins and future

mattrunchey:
## Overview

Given the move to py3, I went through every single package and file (source code, not tests) and cleaned up what I saw. The root goal was to get rid of past and future and monotonic, but I took it a step further and cleaned up the entire codebase to some extent. 

## Major Changes

- All past.builtins and future references removed, as well as __future__ imports at the top of files
- Removed (most) class definitions that inherited from object
- Fixed a bunch of PEP8 complaints
- Amend pylintrc to check for the object inheritance
- Remove monotonic (replace with the import `from time import monotonic`)
- Bump releases so that we can sever dependency on those 3 packages on latest.

### Additional Notes

- We have a lot of source code.


mattrunchey:
OK. Tests passing. All known references to __future__, future, past, and monotonic imports gone. pytest installation shows no evidence that future and monotonic were installed. Will need to bump versions to release.

mattrunchey:
Just kidding. We have *one* dependency on `future` from `jlink` by way of `pylink-square`...

```
Collecting future (from pylink-square>=0.1.3->iotile-transport-jlink==1.0.2)
```

PR #725, opened 2019-03-06
Tag was reused, need 3.0.3 bump

mattrunchey:
I just learned that you can't just create a new tag with the same master target on the repo, even if the pypi package wasn't released it still uses what the tag was originally created with (probably a good idea on their part). However, this means I do need to move it to 3.0.3

### DIFF
--- a/iotilebuild/RELEASE.md
+++ b/iotilebuild/RELEASE.md
@@ -2,13 +2,9 @@
 
 All major changes in each released version of IOTileBuild are listed here.
 
-## HEAD
+## 3.0.3
 
 - Remove past/future/monotonic, pylint cleanup
-
-## 3.0.2
-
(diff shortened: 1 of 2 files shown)


timburke:
This shouldn't be required.  I've reused tags many times previously.  If you delete the release and the tag from GitHub, it's gone forever and then you are free to create another tag with the same name.  I've done this a lot from the command line and I wouldn't expect the gui to behave differently but maybe?

mattrunchey:
I thought the same but the tag got created 4 commits back from master, so the travis CI build didn't pick up that I'd added the 3.0.2 on the build...

mattrunchey:
Not needed; I forgot to delete the tag AND the release so it was using old commits.

PR #731, opened 2019-03-12
Add the ability to completely undefine a -D symbol.

### LINKED ISSUE #730: There should be a way to #undefine a symbol in an architecture
Currently if you have an architecture that declares a CPP `#define` like `FOO=1`, you cannot undefine that definition in another architecture.  You can overwrite the value but you can't make the definition go away.  This is bad in cases where the `#define` is used by 3rd party code and the check is `#ifdef` instead of `#if`.  In that case even redefining the system to have the numerical value 0 will still pass the `#ifdef` check.

timburke:
This PR adds the ability to completely remove a `defines` entry from a previous architecture in a subsequent one.  Previously it was only possible redefine it's value, not `#undefine` it entirely.  Complete removal is important when the `#define` is used by third party code that checks for its presence using `#ifdef` rather than `#if`.

Closes #730

### DIFF
--- a/iotilebuild/RELEASE.md
+++ b/iotilebuild/RELEASE.md
@@ -2,6 +2,13 @@
 
 All major changes in each released version of IOTileBuild are listed here.
 
+## 3.0.5
+
+- Add support for completely undefining a symbol previously included in an
+  architecture's `defines` key in a subsequent architecture.  Now if a "defines"
+  key has the value None (null in json), it will be ignored and not passed 
+  as a -D flag to the compiler.  (Issue #730)
+
 ## 3.0.4
 
 - Change gcc calling routine to use command files (@path_to_args_file) rather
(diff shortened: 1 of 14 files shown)


timburke:
> One note is that the test artifacts for this seem pretty complex (several files, long linker). I assume this is the easiest way to test this functionality, but we should make sure that there's enough context so that when looking back at this months or years later, we know what all of the test artifacts are used for in case we need to change them.

Agreed.  Unfortunately, the easiest way to test this is just with an e2e test of `iotile build` which needs a complete component folder structure to run against.  In this case all of the test artifacts are in a single folder that is used only for this test and referenced together so it should be pretty clear they go together:
https://github.com/iotile/coretools/pull/731/files#diff-a68e17413d43cfb34cbd9edf630d90a2R226

PR #732, opened 2019-03-14
Migrate the ServiceStatusClient to use asynchronous behavior.

mattrunchey:
Since there aren't users of this function outside of the gateway, this change is self contained and avoids prematurely infecting the larger code base with asyncio. However, there is still another dependence on the ServiceStatusClient, so the new functionality has been implemented with a separate Async* naming convention. The long term goal is to deprecate the non-async code paths. 

## Overview

This is a cleaner version of the other WIP PR that cuts out the dev work that explored how deep asyncio might go in the code.

## Major Changes

- An AsyncServiceStatusClient
- An AsyncValidatingWSClient
- An EventLoop utility that enables modules to unify coroutines in a singular event loop that exists in an EventLoopThread for a given program runtime.
- Modification to the virtual_tile to wrap a coroutine-enabled ServiceStatusClient's RPC method for a non-async virtual device.

### Additional Notes

- There are two new test modules that exercise this functionality in parallel to the old tests. Note that it leverages `concurrent.Futures` to schedule coroutines, but since the EventLoop doesn't guarantee ordering of submitted tasks, sometimes there are enforced sleeps to ensure that the async objects have had time to initialize fully.

### DIFF
--- a/iotilecore/iotile/core/utilities/event_loop.py
+++ b/iotilecore/iotile/core/utilities/event_loop.py
@@ -1,0 +1,105 @@
+"""A class that lets you push tasks to a global event loop of the program, letting
+your objects create a reference to the asyncio event loop withotu having to
+pass a loop everywhere
+"""
+
+import asyncio
+import logging
+import threading
+
+
+class EventLoop:
+    """EventLoop
+
+    A container for the asyncio event loop in a background thread
+    """
+
+    loop = None
+    thread = None
+
+    _started = False
+
+    _logger = logging.getLogger(__name__)
+
+    @classmethod
+    def start(cls):
+        """Starts the loop instance if not already created and started"""
+        if not cls.loop:
+            cls._logger.debug("starting event loop")
+            cls.loop = asyncio.new_event_loop()
+            cls.thread = threading.Thread(target=cls._loop_thread_main, name="EventLoopThread", daemon=True)
+            cls.thread.start()
+            cls._started = True
+
+    @classmethod
+    def get_loop(cls):
+        """Loop getter for users of the event loop object"""
+        if not cls.loop:
+            cls.start()
+
+        return cls.loop
+
+    @classmethod
+    def get_thread(cls):
+        """Utility function to get the loop's thread ID for lower-level control/access"""
+        return cls.thread
+
+    @classmethod
+    def stop_loop_abrupt(cls):
+        """Utility function to stop without trying to clean up"""
+        if cls.loop and cls.loop.is_running():
+            cls.loop.stop()
+
+    @classmethod
+    def stop_loop_clean(cls):
+        """Proper way to close the event loop that attempts to clean up all tasks"""
+        if cls.loop and cls.loop.is_running():
+            tasks = asyncio.Task.all_tasks(loop=cls.loop)
+            cls.loop.call_soon_threadsafe(cls.loop.create_task, cls._clean_tasks(tasks))
+            cls.thread.join()
+            cls.loop.stop()
+
+    @classmethod
+    def close(cls):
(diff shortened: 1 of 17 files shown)


mattrunchey:
Tests passing now; there seems to be a flaky test somewhere in ws that I've seen before in normal use that I should file an issue for though.

iotilegateway/iotilegateway/supervisor/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .status_client import ServiceStatusClient
+from .status_client import ServiceStatusClient, AsyncServiceStatusClient
 from .supervisor import IOTileSupervisor
 
-__all__ = ['ServiceStatusClient', 'IOTileSupervisor']
+__all__ = ['ServiceStatusClient', 'AsyncServiceStatusClient' 'IOTileSupervisor']
```
timburke:
```suggestion
__all__ = ['ServiceStatusClient', 'AsyncServiceStatusClient', 'IOTileSupervisor']
```

iotilecore/iotile/core/utilities/event_loop.py
```diff
@@ -0,0 +1,105 @@
...
+
+    _started = False
+
+    _logger = logging.getLogger(__name__)
+
+    @classmethod
+    def start(cls):
+        """Starts the loop instance if not already created and started"""
+        if not cls.loop:
+            print("starting event loop")
```
timburke: I believe this should probably be a logging statement instead.

iotilegateway/iotilegateway/supervisor/status_client.py
```diff
@@ -5,11 +5,490 @@
...
 from time import monotonic
 from iotile.core.hw.virtual import RPCInvalidArgumentsError, RPCInvalidReturnValueError
-from iotile.core.utilities.validating_wsclient import ValidatingWSClient
+from iotile.core.utilities.validating_wsclient import ValidatingWSClient, AsyncValidatingWSClient
+from iotile.core.utilities.event_loop import EventLoop
 from iotile.core.exceptions import ArgumentError
 from . import command_formats
 from . import states
 
+import asyncio
```
timburke: I believe this should be grouped with the other stdlib imports

iotilecore/iotile/core/utilities/event_loop.py
```diff
@@ -0,0 +1,105 @@
...
+    def _loop_thread_main(cls):
+        """This is the background thread that actually owns the single asyncio event loop that your
+        application should be running.
+        """
+        try:
+            cls.loop.run_forever()
+        except:
+            cls._logger.exception("Exception raised from event loop thread")
+        finally:
+            print("stopping loop clean")
```
timburke: This should be a logging statement correct?

iotilegateway/iotilegateway/supervisor/service_tile.py
```diff
@@ -65,7 +69,13 @@ def call_rpc(self, rpc_id, payload=bytes()):
         # run time of RPCs that could be long running.  The caller of the RPC
         # through the tile will know what an appropriate timeout is for the
         # RPC that they are trying to call.
-        resp = self._client.send_rpc(self._service, rpc_id, payload, timeout=120.0)
+        if inspect.iscoroutinefunction(self._client.send_rpc):
```
timburke: I worry about leaving this in here since we ended up not deciding to use the asyncio event loop for this tile at this point.  I would rather leave this tile as it is because when we come back and port over to asyncio for real it will likely be with a different base class like `AsyncVirtualTile` where `call_rpc` itself is a coroutine so this isn't needed.

PR #734, opened 2019-03-19
Update UTCAssigner to be faster and more correct.

### LINKED ISSUE #733: Update UTCAssigner to search in both directions when fixing timestamps
Currently the `UTCAssigner` fixes timestamps to UTC by looking for know anchor points in the in the data stream where it knows a `reading_id` and a UTC time.  It then assigns utc times to all other points by searching for the nearest anchor point and calculating the time offset between that reading and the anchor.

However, it only searches for anchor points that are **after** the given reading that it is searching for.  It should be updated to search in both directions and pick the "best" one based on which direction is more exact, letting the user indicate a preference for which direction to use if both are equally good.

timburke:
This commit changes UTCAssigner to memoize anchor points to reduce the amount of work it needs to do to fix an entire report.

It also updates the utc assignment algorithm to look for reference times in both directions (forward and backward in time) from a given point that it wants to assign a timestamp to.  These changes are designed to make `UTCAssigner` more robust in practice when working with POD-1M data that has a known anchor point before any data is recorded (the trip start event) as well as general PODs that use the received_time of the report itself as their only anchor point.

Closes #733


iotilecore/RELEASE.md
```diff
@@ -2,6 +2,20 @@
...
+
+- Update UTCAssigner with more test coverage and add fix_report() method that
+  will fix all readings inside of a SignedListReport with memoization to 
+  speed up the fixing process. 
+
+- Update UTCAssigner logic to check both directions from a reading to find out
+  which part produces a more exact UTC timestamp, choosing the best one
+  automatically.
+
+- Update `IOTileReading` to do a better job of assigning itself a debug utc
```
mattrunchey:
```suggestion
- Update `IOTileReading` to do a better job of assigning itself a debug UTC
```

PR #738, opened 2019-03-23
Feat event loop

timburke:
## Overview

This PR refactors EventLoop to add support for stopping background tasks in an orderly fashion and adds basic test coverage of the task creation and shutdown process.  For increased testability, `EventLoop` is moved from class methods onto to a normal class (`BackgroundEventLoop`) and a single global instance is created named `EventLoop` for backwards compatibility.

We should probably rename `EventLoop` to something like `SharedLoop` or `GlobalLoop` to better distinguish it from the `BackgroundEventLoop` class.

## Notes

- This PR is to merge into the long-running branch `extended-asyncio-upgrade` that can accumulate asyncio-related changes until we are at a point where we want to merge into `master` and declare that we're going to start maintaining backwards compatibility and release the new functionality.

### DIFF
--- a/iotilecore/iotile/core/utilities/__init__.py
+++ b/iotilecore/iotile/core/utilities/__init__.py
@@ -2,5 +2,6 @@
 
 from .validating_dispatcher import ValidatingDispatcher
 from .workqueue_thread import WorkQueueThread
+from .event_loop import EventLoop, BackgroundEventLoop
 
-__all__ = ['ValidatingDispatcher', 'WorkQueueThread']
+__all__ = ['ValidatingDispatcher', 'WorkQueueThread', 'BackgroundEventLoop', 'EventLoop']
--- a/iotilecore/iotile/core/utilities/event_loop.py
+++ b/iotilecore/iotile/core/utilities/event_loop.py
@@ -159,7 +159,7 @@ class BackgroundTask:
 
         try:
             await asyncio.wait_for(finished, timeout=self._stop_timeout)
-        except asyncio.TimeoutError:
+        except asyncio.TimeoutError as err:
             # See discussion here: https://github.com/python/asyncio/issues/253#issuecomment-120138132
             # This prevents a nuisance log error message, finished is guaranteed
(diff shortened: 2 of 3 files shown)


mattrunchey:
I'm a bit confused on the convention change that will be for ensuring that we only create one EventLoop for a given program with the changes.

Maybe it's using a pattern I'm not familiar with, but if I had two modules that each had the following code, wouldn't it create two Event Loops, since the loop they use isn't a class variable? What is the preferred way of using the same EventLoop without having to pass a BackgroundEventLoop instance between classes?

runner.py
```
import module1
import module2

module1.py
```
from iotile.core.utilities.event_loop import BackgroundEventLoop
loop = BackgroundEventLoop()
loop.start()
```
module2.py
```
from iotile.core.utilities.event_loop import BackgroundEventLoop
loop = BackgroundEventLoop()
loop.start()
```

**Comment by timburke** [comment]
@mattrunchey Your right that directly creating a `BackgroundEventLoop()` would not be implicitly sharable.  To use the shared event loop you would just import and use the single shared `EventLoop` instance instead:
https://github.com/iotile/coretools/pull/738/files#diff-465d2dfaf2dfd7784fb68750fb17b2beR444

It's functionally the same as having everything be a `classmethod` on the BackgroundEventLoop class itself but with the added benefit that you are free to spin-up a non-global BackgroundEventLoop for testing and have it work the same way as the global shared loop you would normally want to use.  

For use in non-testing code, you would probably never directly import or create `BackgroundEventLoop` but rather just import and use `EventLoop`, which I think should be renamed to something like `SharedLoop` or `GlobalLoop` to explicitly connote that it is shared/global.

In your example then:

runner.py
```python
import module1
import module2
```

module1.py
```python
from iotile.core.utilities.event_loop import EventLoop
EventLoop.start()
```

module2.py
```python
from iotile.core.utilities.event_loop import EventLoop
EventLoop.start()
```

mattrunchey:
AH, I think I misinterpreted the pylint comment to mean the creation of that object was for compatibility rather than for python compat.

PR #740, opened 2019-03-25
Refactor asyncio tools into `async_tools` folder and websockets into `websockets` folder

### LINKED ISSUE #741: Rename global EventLoop to SharedLoop or GlobalLoop
It's confusing to have a shared instance named `EventLoop` because it makes you think like maybe its not shared and wonder what the difference is between `EventLoop` and the class `BackgroundEventLoop`.  

I would propose renaming `EventLoop` to `SharedLoop`.

timburke:
## Overview

This PR does three main things:

1. It moves `asyncio` related utilities into `utilities.async_tools` since we will be gathering a decent number of them.  This is the place for any general purpose utility class built for `asyncio` use like `BackgroundEventLoop` or the new `OperationManager`.

2. It adds `OperationManager` which is a message handling mechanism that allows coroutines to cleanly block until receiving one or more specific messages.  Right now it is just used to simplify the implementation of `AsyncValidatingWSClient` but its target use case is refactoring complex `DeviceAdapter` classes and simplifying their implementations.

3. It moves websocket related functionality to `utilities.websockets`.  It refactors `AsyncValidatingWSClient` to use `OperationManager` and adds an initial implementation of `AsyncValidatingWSServer` to simplify the creation of websocket servers that can be interacted with using `AsyncValidatingWSClient`.

## Minor Changes

- A `log_coroutine` method is added to `BackgroundEventLoop`.  This allows you to "fire and forget" coroutines whose continued existence you don't want to track while still logging errors if they raise an exception and not getting a "Future result not inspected" messagewhen shutting down the event loop.
- Fixed two old unit tests for StoppableWorkerThread that apparently created two background threads that pegged the CPU to 100% and never shut them down.

Closes #741


iotilecore/iotile/core/utilities/async_tools/operation_manager.py
```diff
@@ -0,0 +1,318 @@
+"""Message driven asynchronous state machines.
+
+This module allows you to define a complex operation that blocks at certain
+points to wait for one or more messages to be received.  There can be
+many operations running in parallel and multiple operations can be triggered
+by the same message.  Messages are just dictionaries and the matching can
+be according to any keys in the messages.
+
+The Operationmanager sclass is designed to provide a friendly API on which to
```
timburke:
```suggestion
The OperationManager sclass is designed to provide a friendly API on which to
```

PR #759, opened 2019-03-28
Create modern AbstractDeviceAdapter and port VirtualAdapter to use it

### LINKED ISSUE #746: Add coroutine based version of DeviceAdapter
This will be used as the basis of the refactor of many device adapters to use asyncio rather than multithreading.

### LINKED ISSUE #747: Make AsyncDeviceAdapter usable by AdapterStream
This is necessary for HardwareManager to be able to attach to AsyncDeviceAdapter classes without needing to know anything about asyncio.

timburke:
## Overview

This PR lays the foundation for implementing device adapters using coroutines inside of BackgroundEventLoop.  It demonstrates how to do it by porting `VirtualDeviceAdapter`  over to the new system.  It also includes a shim `SynchronousLegacyWrapper` that can turn a modern asyncio DeviceAdapter into the old callback or blocking form.  This shim is installed into `AdapterStream` and `DeviceManager` so that `HardwareManager` and `iotile-gateway` can use the new adapters and fixes any unit tests that broke with the transition.

## Major Changes

A more granular list of the changes made, ideally with links to issues

- Implement `AbstractDeviceAdapter`.  This is a documented abstract base class for modern coroutine based device adapters.  It has no implementation but documents all of the methods that a device adapter must implement and what the requirements on the implementation must be.

   The class is designed so that it can encompass all of the functionality in the legacy DeviceAdapter as well as the `iotile-gateway` DeviceManager class so that DeviceManager is just another standard device adapter.  
- Implement `SynchronousLegacyWrapper`.  This is a wrapper class that provides the legacy API on top of a new AbstractDeviceAdatper to that it can be used in CoreTools as is.  This wrapper is automatically used inside `DeviceManager` and `AdapterStream` so that both modern and legacy device adapters can now be used.
- Implement `StandardDeviceAdapter`.  This is a standard implementation of the `AbstractDeviceAdapter` interface that includes reference implementations of per-connection data tracking and event notification.  It is expected that most real device adapters inherit from this class class unless they need something crazy or unusual that makes them not want the standard implementations.
- Port `VirtualDeviceAdapter` to inherit from `StandardDeviceAdapter` as a test case of how to port a legacy device adapter to the new format.  A minor tweak to `EmulatedDeviceAdapter` was needed since that inherited from `VirtualDeviceAdapter`.  Minor tweaks on a few unit tests were needed that directly assumed the implementation details of `VirtualDeviceAdapter` but mostly it just worked.

### Additional Notes

- Refactors `adapter.py` into a subpackage named adapter to be able to reorganize the legacy vs modern code.
- Adds a few more convenience routines to `BackgroundEventLoop` to make subtask management easier.
- Refactors websockets `async-server.py` to take advantage of new subtask functionality.

Closes #746 
Closes #747


timburke:
@mattrunchey BTW, I know this is too large to effectively review but it was the smallest group of things I could put together that coherently let me go on to build actual device adapters and ensure that the base was solid enough that we would be able to remove a lot of code in the future.  

As I started refactoring the websockets device adapter I was realizing that unless we were careful in the design of `AbstractDeviceAdapter` we wouldn't get as much simplification as we could otherwise achieve so I paused and used that to design a solid, modern DeviceAdapter interface using all of the benefits we can get from coroutines and BackgroundEventLoop and am now going back and cleaning up the websockets adapter based on this new foundation.

timburke:
I also wanted to make sure that none of the unit tests broke until the complete refactor was done, which necessitated putting shims in place in the same PR so that everything still works as the refactor goes forward.

PR #765, opened 2019-03-29
Add generic device server implementation and refactor progress notifications

### LINKED ISSUE #754: Unify VirtualInterface and GatewayAgent
There is no longer a need to have a distinct VirtualInterface and GatewayAgent since both now have access to an event loop.

### LINKED ISSUE #756: Document event object format for all events raised by AbstractDeviceAdapter
We need to properly document what kind of object must be sent in each type of event and what attributes or keys it must contain for interoperability.  

There are a lot of implicit rules about scan payloads for example that we should collect and document in terms of how the expiration time needs to be reported.

timburke:
## Overview

Adds  `AbstractDeviceServer` and `StandardDeviceServer` classes.  These classes are designed to unify and eventually replace `VirtualInterface` and `GatewayAgent`.  

`AbstractDeviceServer` is designed to let `iotile-gateway` load and run device servers without needing to know anything about how they serve devices.  It is intentionally left very minimal.

`StandardDeviceServer` is designed to be the reference implementation of device server functionality that most/all device servers inherit from.  As described in the documentation, it takes a single `AbstractDeviceAdapter` instance and serves multiple clients access to the devices that are reachable through that adapter.

`StandardDeviceServer` takes care of all of the tricky house-keeping around cleaning up after clients that unexpectedly go away and leave device connections hanging.  The goal is to have all that is required to implement a new DeviceServer be just listening for connections (over http, tcp, etc) and translating the messages received into calls on `StandardDeviceServer`.

## Major Changes

- Add `AbstractDeviceServer` class
- Add `StandardDeviceServer` class that implements `AbstractDeviceServer`
- Refactor how progress notifications are sent to not be callback based but to use the same notification system as other events.

### Additional Notes

- Slightly reorganizes the `transport` subpackage to be more logical.  
- Updates `SynchronousLegacyWrapper` to explicitly inherit from DeviceAdapter and provide both `_sync` and `_async` versions of all legacy functions.  It takes care of translating the new `progress` events back to the old progress callbacks.
- Slight refactor on `BackgroundEventLoop.launch_coroutine` to allow it to be called from within and without of the loop itself.  It now returns an `asyncio.Future` or `concurrent.futures.Future` based on the calling thread.  This lets us simplify the implementation of `notify_events`.
- Documents all of the events that can be notified by an AbstractDeviceAdapter.

Closes #756
Closes #754

### DIFF
--- a/iotilecore/iotile/core/hw/transport/adapter/abstract.py
+++ b/iotilecore/iotile/core/hw/transport/adapter/abstract.py
@@ -506,7 +506,7 @@ class AbstractDeviceAdapter(abc.ABC):
             RPCErrorCode: The RPC implementation wishes to fail with a
                 non-zero status code.
             RPCInvalidIDError: The rpc_id is too large to fit in 16-bits.
-            TileBusSerror: The tile was busy and could not respond to the RPC.
+            TileBusyError: The tile was busy and could not respond to the RPC.
             Exception: The rpc raised an exception during processing.
             DeviceAdapterError: If there is a hardware or communication issue
                 invoking the RPC.
--- a/iotilecore/iotile/core/hw/transport/adapter/abstract.py
+++ b/iotilecore/iotile/core/hw/transport/adapter/abstract.py
@@ -26,6 +26,89 @@ DeviceAdapter.
 Previous DeviceAdapter implementations used multithreading and callback based
 interfaces that worked but proved difficult to implement, raising the barrier
 to creating new DeviceAdapters to support additional communication protocols.
+
+
+Events
+======
+
+Communication with a device through an AbstractDeviceAdapter is bidrectional
+in that commands can flow from the user to the device and events can flow
+back from the devie to the user without being the direct result of a command.
+
+The way to receive events is to register an event monitor using
+:meth:`AbstractDeviceAdapter.register_monitor`.  There are different kinds of
+events that you can receive and your monitor includes filters based on the
+event name and the source device.
+
+Every event notification contains a tuple with three things: the connection
+string of the device that sent the event, the name of the event and an event
+object whose contents vary depending on the event name.  The supported events
+are:
+
+    report
+        A report has been received from the device.
+
+        The event object will be a subclass of
+        :class:`iotile.core.hw.reports.IOTileReport`.
+        This event is used to send reports from a device to the user via the
+        streaming interface.  It can only be received if the streaming
+        interface is open.
+
+    connection
+        Someone has connected to a device.
+
+        The event object for this event will be ``None`` since there is not
+        additional information available.
+
+    trace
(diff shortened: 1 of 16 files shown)


iotilecore/iotile/core/hw/transport/adapter/abstract.py
```diff
@@ -50,14 +133,15 @@ def can_connect(self):
...
-        because a user can just try to connect to a device and have it fail if
-        the adapter cannot accomodate another connection.  However, in more
-        complicated scenarios where there are multiple DeviceAdapters that
-        could be used to connect to a device, it is useful to be able to
-        filter them and find the best adapter that has a free connection
-        slot available.
+        given DeviceAdapter may be able to maintain simultaneous connections
+        to more than one device at a time.  In simple cases, this method is
+        not necessary because a user can just try to connect to a device and
+        have it fail if the adapter cannot accomodate another connection.
```
mattrunchey:
```suggestion
        have it fail if the adapter cannot accommodate another connection.
```

iotilecore/iotile/core/hw/transport/adapter/abstract.py
```diff
@@ -419,18 +506,19 @@ def remove_monitor(self, handle):
             RPCErrorCode: The RPC implementation wishes to fail with a
                 non-zero status code.
             RPCInvalidIDError: The rpc_id is too large to fit in 16-bits.
+            TileBusSerror: The tile was busy and could not respond to the RPC.
```
mattrunchey: Is there a typo here? That "S" sticks out

iotilecore/iotile/core/hw/transport/adapter/mixin_notifications.py
```diff
@@ -190,8 +196,35 @@ def remove_monitor(self, handle):
...
+        """Notify an event.
+
+        This will move the notification to the background event loop and
+        return immediately.  It is useful for situations where you cannot
+        await notify_event but keep in mind that it prevents back-pressure
+        when you are notifying too fast so should be used sparingly.
+
+        Note that calling this method will push the notification to a
+        background task so it can be difficult to reason about when it will
+        precisely occur.  For that reason, :meth:`notify_event` should be
```
mattrunchey: The language here suggests that we should prefer an alternate method, but it names the same method.

iotilecore/iotile/core/hw/transport/adapter/abstract.py
timburke: yeah, `TileBusyError`

iotilecore/iotile/core/hw/transport/adapter/mixin_notifications.py
timburke: Good catch, updating the docstring.

PR #769, opened 2019-04-01
Feat websockets

### LINKED ISSUE #744: Reimplement Websockets DeviceAdapter on top of AsyncValidatingWSClient

### LINKED ISSUE #745: Reimplement Websockets GatewayAgent on top of AsyncValidatingWSServer

### LINKED ISSUE #749: Reimplement WebsocketVirtualInterface using AsyncValidatingWSServer
Now that we have asyncio, this VirtualInterface can be unified on top of the GatewayAgent without needing a fully separate class since the GatewayAgent will no longer depend on tornado.

### LINKED ISSUE #761: Move all websockets functionality to iotile-transport-websockets
Right now there is a bunch of websockets related functionality inside `iotile-core` that make it easy to create websockets clients and servers but this functionality is only used inside `iotile-gateway` and `iotile-transport-websockets`. 

It makes more sense to refactor dependencies as follows:

- `iotile-core`.  Nothing related to websockets
- `iotile-transport-websockets` Both the underlying client/server implementations as well as a GatewayAgent and a DeviceAdapter that use that functionality.
- `iotile-gateway` depends on `iotile-core` and `iotile-transport-websockets` so that it can ship with a default websockets transport protocol and so that `iotile-supervisor` can use the asyncio server feature from `iotile-transport-websockets`.

Basically this means moving `async_client` and `async_server` to `iotile-transport-websockets` and then updating `iotile-gateway` to find it there rather than from `iotile-core`.

timburke:
## Overview

Moves all websockets functionality from iotile-core to `iotile-transport-websocket` and implements `DeviceAdapter` and `DeviceServer` for websockets, testing them against each other.  Also ports `WebsocketVirtualInterface` to be a thin shim on top of `WebsocketsDeviceServer`.  This will eventually go away once we remove the need for `VirtualInterface` and allow the `virtual_device` script to directly use a `DeviceServer`.

## Major Changes

- Implement `WebsocketDeviceAdapter`.  This is the new `asyncio` based websocket device adapter built on top of `AsyncValidatingWSClient` and `StandardDeviceAdapter`.  Everything except forwarding disconnections and debug commands is implemented and tested.
- Implement `WebsocketDeviceServer`.  This is a unification of `WebsocketVirtualInterface` and `WebsocketGatewayAgent`
- Removed old virtual interface and turn into a thin shim on top of `WebsocketDeviceServer`
- Moved generic websocket code here from `iotile-core`.

### Additional Notes

- Updated `OperationManager` to allow coroutine based callbacks so that we can guarantee that notifications are received and processed at a known point in time.

Closes #761
Closes #749
Closes #745
Closes #744


iotilecore/iotile/core/hw/virtual/__init__.py
```diff
@@ -7,11 +7,15 @@
                            RPCNotFoundError, RPCInvalidArgumentsError,
                            RPCInvalidReturnValueError, TileNotFoundError,
                            RPCErrorCode, BusyRPCResponse, unpack_rpc_payload, pack_rpc_payload,
-                           pack_rpc_response)
+                           pack_rpc_response, unpack_rpc_response,
+                           VALID_RPC_EXCEPTIONS)
+
 from .virtualinterface import VirtualIOTileInterface
 
 __all__ = ['VirtualTile', 'VirtualIOTileDevice', 'tile_rpc',
```
mattrunchey: It seems like we're collecting quite the list of imports here. Is this a normal thing or do alternative solutions exist for situations like this? I tried briefly to see if anyone posted thoughts about this on the web without success. 

Not really important, just a thought.

timburke: Yeah we should tag this for a refactor to see if these are all still required or if there is a way we can group some higher level functionality that hides some of them.

timburke: Tagged as #771

PR #772, opened 2019-04-02
Port iotile-supervisor to asyncio completely

### LINKED ISSUE #742: Refactor AsyncSupervisorStatusClient on top of new AsyncValidatingWSClient
Should be a very easy port but some method use may need to be renamed.  This should be a test case of whether the API of AsyncValidatingWSClient is good or should be adjusted based on how easy it is to subclass it.

### LINKED ISSUE #743: Reimplement iotile-supervisor server on top of AsyncValidatingWSServer
This can be a test case for the useability of the new `AsyncValidatingWSServer` implementation and show us how easy it will be to complete the move off of tornado.

timburke:
## Overview

This PR completely moves IOTileSupervisor from tornado to `asyncio` using BackgroundEventLoop.  The new supervisor is built on top of `AsyncValidatingWSServer` just like the client is built on `AsyncValidatingWSClient`.  The communication protocol for RPCs is changed slightly to fit with how those two base classes work.  All of the unit tests are ported over as well and the command line wrappers are updated to work on python 3 and remove any dependence on tornado.

## Major Changes

- Move `IOTileSupervisor` from tornado to `AsyncValidatingWSServer`.
- Move `iotile-supervisor` script to no longer need tornado.  Adds additional command line option for controlling log verbosity.
- Refactor `AsyncSupervisorStatusClient` to use new `AsyncValidatingWSClient` functionality.
- Refactor `SupervisorStatusClient` to just be a thin wrapper around `AsyncSupervisorStatusClient`.
- Rename `SupervisorStatusClient` to `SupervisorClient` and `AsyncSupervisorClient` since they do way more than just sync status information now.
- Update all of the unit tests
- `iotile-gateway` package now depends on `iotile-transport-websocket` since it uses the generic websocket client/server functionality from that package.  

### Additional Notes

- Adds `AwaitableDict` utility class to facilitate blocking on a future event identified by a unique key.  This is used to signal when an RPC has been finished.
- Refactors `ServiceManager` notifications to allow coroutines so that we can forward them over websockets.
- Updates `OperationManager` to allow `process_message` to block until the message has been processed through all callbacks to be able to ensure the order of notifications being sent.
- `iotile-supervisor` script could not previously load json config on python 3.  That is fixed now.

### Future Plans

The next step is to port `iotile-gateway` itself off of tornado and onto `AsyncValidatingWSServer` and then we can officially drop tornado and all of the `ws4py` functionality.  That will be the next PR.

Closes #742 
Closes #743


iotilecore/iotile/core/utilities/async_tools/awaitable_dict.py
```diff
@@ -0,0 +1,159 @@
...
+        """Get the value of a key if it is already set.
+
+        This method allows you to check if a key has already been set
+        without blocking.  If the key has not been set you will get the
+        default value you pass in or KeyError() if no default is passed.
+
+        When this method returns the key is automatically removed unless
+        you pass ``autoremove=False``.
+
+        This method is no a coroutine and does not block.
```
mattrunchey:
```suggestion
        This method is not a coroutine and does not block.
```

iotilecore/iotile/core/utilities/async_tools/event_loop.py
```diff
@@ -262,6 +263,37 @@ def start(self):
             self.thread = threading.Thread(target=self._loop_thread_main, name="EventLoopThread", daemon=True)
             self.thread.start()
 
+    def wait_for_interrupt(self, check_interval=1.0, max_time=None):
+        """Run the event loop until we receive a ctrl-c interrupt or max_time passes.
+
+        This method will wake up every .1 seconds by default to check for
```
mattrunchey: Seems to be some discrepancy in the default by the comment (0.1 seconds) and the default arg (1 second)?

timburke: nice catch, fixed.

iotilecore/iotile/core/utilities/async_tools/awaitable_dict.py
timburke: Fixed in same new commit

PR #782, opened 2019-04-08
Finalize support for coroutine based device adapters in HardwareManager

### LINKED ISSUE #768: Cleanup and refactor hw.exceptions hierarchy
Most of the exceptions in `iotile.core.hw.exceptions` are no longer relevant and many have been superseded by classed defined in `iotile.core.hw.virtual` but the older ones have not been audited for usage across the code base and removed.

### LINKED ISSUE #778: Refactor AdapterCMDStream to natively work with AbstractDeviceAdapter
Now that we have a modern DeviceAdapter implementation and a shim wrapper that can provide the same interface on top of legacy device adapters, we should update AdapterCMDStream to natively work with the new class and wrap the legacy class automatically. 

We should also clean it up and remove the useless `CMDStream` class once and for all.

### LINKED ISSUE #779: Remove duplicate hardware errors
There are some _very legacy_ exceptions like `ModuleNotFoundError` that are only used to communicate between `proxy.py` and `cmdstream.py`.  We since created a modern hardware error hierarchy that has duplicate errors but never cleaned up and removed the old ones.  

We don't actually use the old errors anywhere outside of CoreTools:
https://github.com/search?q=org%3Aiotile+ModuleNotFoundError&type=Code

timburke:
## Overview

This PR finalizes the `asyncio` port of CoreTools with the last major bit of functionality needed for an `asyncio`-first device transport layer.  It also takes the opportunity to clean up a lot of the very old code around `AdapterCMDStream` and `CMDStream` to modernize it.  

## Major Changes

1. Refactor `AdapterCMDStream` to make it natively use `AbstractDeviceAdapter` instead of working through a shim.  Adds a shim to allow the new `AdapterStream` to work with legacy device adapters by "forward-porting" them to have a coroutine based API.
2. Cleanup duplicate exceptions between `iotile.core.hw.exceptions` and `iotile.core.hw.virtual`.  Now all exceptions live in `iotile.core.hw.exceptions`, are properly documented and used consistently everywhere.
3. Updates the `enable_*` functionality in `AdapterStream` and `HardwareManager` to support opening an interface multiple times cleanly.
4. Completely removes `CMDStream` class that doesn't do anything anymore and merges its functionality into `AdapterStream`.
5. Removes support for creating `CMDStream` subclasses inside plugins.  This was only ever used by the websocket stream that we replaced with a proper device adapter anyway.  There is also no longer a need for CMDStream subclasses because the `AbstractDeviceAdapter` interface is clean enough that it can be just implemented directly rather than needing an "easier" interface for basic usage.
6. Improves documentation wherever possible and removes some legacy routines from HardwareManager that haven't worked in a number of years.
7. Refactors `proxy.py` to remove duplicate code between `rpc` and `rpc_v2` and use the new exception hierarchy.
8. Adds documented way for `SharedLoop` to stop itself at process exit and removes ancient hack to have `atexit` routines registered for each `CMDStream` that was open.  Now `SharedLoop` cleans up all `AdapterStreams` just like normal background tasks.
9. Removes an old hack in the way the `jlink` debug functionality was implemented that broke the fundamental invariant that the way you use an iotile device is "connect", "open interface", "use interface".  The jlink plugin allowed you to "open" the `debug` interface on a device without first connecting to that device.  This has been removed but the jlink plugin will need to be updated to correctly implement the `debug` interface.  I'm not including that in this PR since the jlink plugin is broken on python 3 currently anyway and needs its own fixes.

### Testing Done

1. All controller hardware RPC based unit tests run and pass with the new code.
2. Reflashing devices and debugging them using `DebugManager` works except for known python 3 bugs in the `jlink` transport plugin.

Closes #779 
Closes #778 
Closes #768


iotilecore/iotile/core/hw/exceptions.py
```diff
@@ -70,48 +69,119 @@ def __init__(self, client_id, conn_str, operation, reason):
...
+    """
+
+    def __init__(self):
+        super(AsynchronousRPCResponse, self).__init__("RPC handler elected to return asynchronously")
+
+
+class BusyRPCResponse(RPCError):
+    """The Tile was busy and could not handle an RPC right now.
+
+    This exception is rarely exposed diretly to a user.  Internally, it can be
```
mattrunchey:
```suggestion
    This exception is rarely exposed directly to a user.  Internally, it can be
```

iotilecore/iotile/core/hw/transport/adapterstream.py
```diff
@@ -85,200 +121,450 @@ def _scan(self, wait=None):
...
+            connstring = self._scanned_devices[uuid_value]['connection_string']
+
+        self.connect_direct(connstring)
 
-        connstring = self._scanned_devices[uuid_value]['connection_string']
-        self._connect_direct(connstring)
+    def connect_direct(self, connection_string, no_rpc=False, force=False):
+        """Directly connect to a device using its stream specific connection string.
+
+        Normally, all connections to a deviec include opening the RPC
```
mattrunchey:
```suggestion
        Normally, all connections to a device include opening the RPC
```

iotilecore/iotile/core/hw/transport/adapterstream.py
```diff
@@ -85,200 +121,450 @@ def _scan(self, wait=None):
...
 
-        self.adapter.periodic_callback()
+        return unpack_rpc_response(status, payload, rpc_id, address)
+
+    def send_highspeed(self, data, progress_callback):
+        """Send a script to a device at highspeed, reporting progress.
 
-        if not success:
-            raise HardwareError("Could not send RPC", reason=result['failure_reason'])
+        This method takes a binary blog and downloads it to the device as fast
```
mattrunchey:
```suggestion
        This method takes a binary blob and downloads it to the device as fast
```

PR #791, opened 2019-04-12
Major upgrade to use asyncio instead of multithreading for parallelism

timburke:
## Overview

This PR merges a large set of changes that significantly refactor that `DeviceAdapter` system that is used to find and connect to iotile devices to be based primarily on `asyncio` and coroutines instead of multithreading.  There are shims in place to allow all current transport plugins to work while they are upgraded to be `asyncio` based.

## Major Changes

- Completely Refactors `DeviceAdapter` into `AbstractDeviceAdapter` and `StandardDeviceAdapter`.
- Reimplements `VirtualAdapter` and `WebsocketAdapter` on top of `AbstractDeviceAdapter`.
- Unifies `VirtualInterface` and `GatewayAgent` into a generic `AbstractDeviceServer`.  Adds `StandardDeviceServer` to make it easier to build a device server.
- Removes dependencies on `ws4py` and `tornado` completely.  

### Additional Notes

- These are partially backwards incompatible changes in `iotile-core` and `iotile-gateway` as well as moving some functionality from `iotile-core` into `iotile-transport-websocket`.  This will result in a major version bump of those packages but there needs to be a few additional features added first before we close the "backwards incompatibility" window.


mattrunchey:
Do you typically adjust the `version.py` and such for packages after review? I noticed `iotile-core` is still on v4 and I would imagine this pushes it to v5.0.0-{prerelease}, along with the other packages

timburke:
The remaining work that needs to be added before we cut a release is tracked in the `asyncio` project:

https://github.com/iotile/coretools/projects/5

The primary goal is to continue the cleanup backwards incompatible cleanup of
`VitualIOTileDevice` to make it natively `asyncio` based and remove `VirtualIOTileInterface` entirely, replacing it with `AbstractDeviceServer`.  This requires updating `iotile-transport-bled112`  to be coroutine based.

PR #794, opened 2019-04-16
Add the FilesystemManagerResource and ModifyJsonStep to iotileship

timothyfroehlich:
This will enable iotileship recipes to modify filesystems and block devices.


iotileship/iotile/ship/resources/filesystem_manager.py
```diff
@@ -0,0 +1,120 @@
...
+    Can work on either a directory or a unmounted block device
+
+    Important: mounting and unmounting of a block device requires the udisksctl utility, which is included with
+    most distributions of Ubuntu
+
+    All Steps using this Resource will use self.root as the root directory to find other files
+
+    Root directory is a pathlib.Path
+
+'''
```
mattrunchey: I think our style guide calls for this not having a blank line after the text.

iotileship/iotile/ship/resources/filesystem_manager.py
```diff
@@ -0,0 +1,120 @@
...
+from iotile.core.exceptions import ArgumentError
+from .shared_resource import SharedResource
+
+RESOURCE_ARG_SCHEMA = DictionaryVerifier(desc="filesystem_manager arguments")
+RESOURCE_ARG_SCHEMA.add_required("res_path", StringVerifier("path to resource to work with. "
+                                                            "Specify file:// for a directory or block:// for an unmounted disk partition"))
+
+
+class FilesystemManagerResource(SharedResource):
+    '''
```
mattrunchey: Style guide has a multi-line comment block with text on the same line. Essentially, if the comment all fits on one line, you end up with `'''this is a comment'''`, but on multiline you end up with
```
'''This is a comment
That contains multiple lines
And is a haiku
'''
```

iotileship/iotile/ship/resources/filesystem_manager.py
```diff
@@ -0,0 +1,120 @@
...
+            try:
+                subprocess.run(['udisksctl', 'help'], stdout=subprocess.DEVNULL, check=True)
+            except subprocess.CalledProcessError:
+                print("Could not find 'udisksctl'. Make sure it's installed")
+                raise
+
+        self.root = None
+        self._unmount_in_close = False
+
+    #If self._resource_loc is mounted, returns pathlib.Path to mount location
```
mattrunchey: There should usually be a space between the `#` and the text of a comment. https://realpython.com/python-pep8/#block-comments

(this comment applies throughout the PR)

iotileship/iotile/ship/resources/filesystem_manager.py
```diff
@@ -0,0 +1,120 @@
...
+        for line in info_proc.stdout.splitlines():
+            if "MountPoints:" in line:
+                split = line.split()
+                if len(split) > 1:
+                    return pathlib.Path(split[1])
+                return None
+
+        raise RuntimeError("udisksctl failed to report MountPoints")
+
+
```
mattrunchey: Should be one line between functions in a file, I think. Does your IDE have a PEP8 checker installed? That might make it easier to write stylistically consistent code, but you could also use `pylint` and apply the `.pylintrc` file in the root of coretools.

iotileship/iotile/ship/resources/filesystem_manager.py
```diff
@@ -0,0 +1,120 @@
...
+                                parameter_name="res_path")
+
+        self._type = self._path.split('://')[0]
+        self._resource_loc = self._path.split('://')[1]
+
+        if self._type not in self._allowed_types:
+            raise ArgumentError("FilesystemManagerResource unknown filesystem type", parameter_name="res_path")
+
+        if self._type == "block" and sys.platform != "linux":
+            raise ArgumentError("block;// paths are only supported on Linux. \
```
timburke: `\` line continuations are not needed in python when you are inside of a `( )` because that implicitly continues across lines.

timothyfroehlich: This is the current thing I'm fighting with, as the atomic moves are ruining the permissions. You've also just reminded me that I forgot to document the permission stuff in the iotile ship repo.

The solution as currently implemented in meta-arch-systems is to generate files with a group with a specific gid, assign that group to the iotile user, and require that the user running the personalize recipe also be a member of a group with the same gid. 

In more detail:
I've got the daemons running as the iotile user, with the iotile user assigned to group iotile (specifically assigned to gid 19057). On my local machine, I made a group named "yoc_iotile" with gid 19057 and added my local account to the group. This gives my machine access to rewrite the file.  There's also the passwd file holding the passwd hash for the iotile user, it should be owned by root, technically, but assigning it to a group that is not assigned to any user should accomplish the same protections. It's writeable by gid 15666 and on my local machine I added my user to a group called "yoc_noaccess" with gid 15666. 

Group ID's are typically between 0 and 1000, with user's primary groups being automatically numbered at 1000 and decreasing (user 2's group is 999). Gids up to 60,000 are commonly supported, so I picked two arbitrary gids in that range to use for the iotile group and the noaccess group.

iotileship/iotile/ship/resources/filesystem_manager.py
timothyfroehlich: The files I've added are currently clean of pylint issues, though I neglected to go review the styleguide. Will do that.

iotileship/iotile/ship/resources/filesystem_manager.py
timburke: @timothyfroehlich If you just file.write() does it maintain the permissions even though the file contents are truncated and overwrritten?  Then I would say that we just enforce that all files are precreated and `iotile-ship` only ever changes their contents.

timothyfroehlich: Also, yes, files will be owned by whoever last modifies the file. Changing the group should handle permissions issues. Changing owners would require the iotileship script be run with sudo which I'm very hesitant to require.

timburke: and then we don't have to worry about matching `gid` or user_id.  I would drop all of the atomic writes then per my other comments since the main benefit of atomic writes is when you are running code unattended in a situation where power could be lost at any time.  That's not the situation when you're running `iotile-ship` offline, on an SD card, during manufacturing time.

timburke: It doesn't seem like the file ownership is modified by a write operation, unless `root` is special:

```shell
timburke@eng2:~$ touch temp
timburke@eng2:~$ sudo -s
[sudo] password for timburke:
root@eng2:~# ls -l
-rw-rw-r-- 1 timburke timburke    0 Apr 16 14:41 temp
root@eng2:~# echo "hello" > temp
root@eng2:~# ls -l
-rw-rw-r-- 1 timburke timburke    6 Apr 16 14:42 temp
root@eng2:~# cat temp
hello
```

I created a file as `timburke` and wrote to it as root without affecting the owner or group.

timburke: Yeah I just verified with another user that wasn't root.  Just doing a write (even a complete truncation) to file as long as you don't destroy it doesn't seem to affect the ownership on linux.

timothyfroehlich: So I believe I've identified an issue in arch-node-tasks either way. While the files were created by yocto with 664 permissions, the atomic-rewrite code in mqtt_bridge was recreating the files with 644 permissions, removing the group's write permissions. This caused subsequent changes to fail. I'm not 100% sure because I saw it happen once, but I think I'm going to be a lot happier if I just get everything to use the same code for atomic moving, which needs to include preserving the gid and permissions on the file. The atomicwrite code does not do this.

So here's a new idea, I could make a new atomic re-write that starts by shutil.copy (or copy2?) the file, which would preserve all permissions including owner. If that works it'd probably be best.

timothyfroehlich: Scratch that...
On POSIX platforms, this means that file owner and group are lost as well as ACLs. On Mac OS, the resource fork and other metadata are not used. This means that resources will be lost and file type and creator codes will not be correct. On Windows, file owners, ACLs and alternate data streams are not copied.

https://docs.python.org/3/library/shutil.html#shutil.chown

timothyfroehlich: Could just read the owner / group of the original file and chown the new one.

timburke: It looks like `bindfs` can be used to obviate the need to have the group ownership of a file be numerically equivalent to the group of a user running the personalization step.  It seems like you can tell bindfs to ignore the actual file ownership bits during the bind mount so that you are not prohibited from modifying the files (provided you don't create new files):

https://unix.stackexchange.com/a/161709

That should solve the need to have a shared group between the SD card in question and your computer running `iotile-ship`.

timburke: Here's a more complete example showing how to even map uid/gid using bindfs so you can create files with the correct uid/gid from any linux computer:
https://unix.stackexchange.com/a/158695

timothyfroehlich: The mounting part would need to be run as root, but if I go back to handle any block device mounting in a shell script, then I should be able to just run the required commands and not the whole thing. Or I can make a sudoers file to install that gives permission to do bind.

timothyfroehlich: I'm dragging bad today and a second espresso hasn't helped. I'll let this settle in my head and work on it more tonight.

timburke: Understood.

To summarize, I would say:

- it looks like having this all work magically without needing someone to do something as root at somepoint isn't feasible because we are working with an external filesystem whose numerical ids are not going to match in general with what is on the user's computer.
- this makes the `block://` option not super useful right now unless it is able to elevate and run as root so that the user can actually edit the files that they are supposed to regardless of what the uid/gid on the actual fs are.
- using `bindfs` we can fix all of the uid/gid issues by making the user mount the fs first using bindfs to make all fixes writable by the current user.  
- then we make sure we don't create any new files (inadvertently via atomicwrites, for example) during the `iotile-ship` execution.

This means the user will need to bind mount the directory they want to work on prior to running iotile-ship but that's easy to document in a README

timothyfroehlich:
This merge is necessary for https://github.com/iotile/production_devices/pull/134

iotileship/iotile/ship/actions/ModifyJsonStep.py
```diff
@@ -0,0 +1,89 @@
+import json
+import copy
+
+from iotile.ship.exceptions import RecipeActionMissingParameter
+
+def modify_dict(data, key, value, create_if_missing=False):
+    """
```
mattrunchey: Style: comment blocks are supposed to start on the same line as the opening quotes (throughout)

iotileship/iotile/ship/actions/ModifyJsonStep.py
```diff
@@ -0,0 +1,89 @@
...
+        (dict): the final modified dict
+    """
+    data_copy = copy.deepcopy(data)
+    key_copy = copy.deepcopy(key)
+
+    delver = data_copy
+    current_key = key_copy
+    last_key = "Root"
+
+    #Dig through the json, setting delver to the dict that contains the last key in "key"
```
mattrunchey: pep8: space between pound sign and start of comment (throughout)

iotileship/iotile/ship/actions/ModifyJsonStep.py
```diff
@@ -0,0 +1,89 @@
...
+
+from iotile.ship.exceptions import RecipeActionMissingParameter
+
+def modify_dict(data, key, value, create_if_missing=False):
+    """
+    Change (or add) a json key/value pair.
+
+    Args:
+        data (dict): The original data. This will not be modified.
+        key (str|list): Either a string or a list of keys and subkeys specifing the key to change
```
mattrunchey: Is there an external constraint that makes us have a multi-type argument to a function? This looks a bit complicated - shouldn't it always be a list, just sometimes with a length of one?

iotileship/iotile/ship/actions/ModifyJsonStep.py
```diff
@@ -0,0 +1,89 @@
...
+
+            last_key = current_key[0]
+            delver = delver[current_key[0]]
+            current_key.pop(0)
+
+        final_key = current_key[0]
+    elif isinstance(current_key, str):
+        final_key = current_key
+
+    if final_key not in delver and not create_if_missing:
```
mattrunchey: Technically, you can have final key not declared if you don't input a string or list to this function, since final_key isn't declared before the if/elif

iotileship/iotile/ship/actions/ModifyJsonStep.py
```diff
@@ -0,0 +1,89 @@
...
+        if args.get('value') is None:
+            raise RecipeActionMissingParameter("ModifyJsonStep Parameter Missing", parameter_name='value')
+        if args.get('path') is None:
+            raise RecipeActionMissingParameter("ModifyJsonStep Parameter Missing", parameter_name='path')
+
+        self._key = args.get('key')
+        self._value = str(args.get('value'))
+        self._path = str(args.get('path'))
+        self._create = bool(args.get('create_if_missing', False))
+
```
mattrunchey: pep8: one line between function definitions within a class

iotileship/test/test_jsonstep.py
```diff
@@ -0,0 +1,79 @@
+import json
+import pytest
+from iotile.ship.actions.ModifyJsonStep import ModifyJsonStep
+
+class mock_resource:
+    def __init__(self, root):
+        self.root = root
+
```
mattrunchey: I'm not sure how strict we are on test code being pep8, but top-level functions/classes should have 2 blank lines between them (vs. one line for class methods)

iotileship/iotile/ship/actions/ModifyJsonStep.py
timothyfroehlich: Yeah, that makes the most sense. I'll change it to only accept lists.

iotileship/iotile/ship/actions/ModifyJsonStep.py
timothyfroehlich: Resolved by only accepting lists

PR #796, opened 2019-04-17
asyncio / websockets small fixes

mattrunchey:
In the EventLoop:
 - Add outcome exception printouts for when cancelled tasks have an exception other than CancelledError to help debug
 - Some minor style changes

In websockets:
 - Switched the datetime decoder to not use bytes, similar to what we'd done in the past for the other websockets datetime things for py3

In gateway:
 - short_name was used when long_name was the variable present and (I believe) intended to be used.

### DIFF
--- a/iotilecore/iotile/core/utilities/async_tools/event_loop.py
+++ b/iotilecore/iotile/core/utilities/async_tools/event_loop.py
@@ -63,7 +63,7 @@ class BackgroundTask:
             If not specified, calling stop() will result in cancel()
             being called on the underlying task.
 
-            The finalizer (or task.cancel()) is always invokved inside
+            The finalizer (or task.cancel()) is always invoked inside
             the event loop.  If finalizer is a coroutine it is awaited.
         loop (BackgroundEventLoop): The background event loop this task
             should run in.  If not specified, it defaults to the global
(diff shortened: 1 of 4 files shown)


iotilecore/iotile/core/utilities/async_tools/event_loop.py
```diff
@@ -203,13 +203,13 @@ def add_subtask(self, subtask):
...
 
         try:
-            await asyncio.wait_for(finished, timeout=self._stop_timeout)
+            outcomes = await asyncio.wait_for(finished, timeout=self._stop_timeout)
         except asyncio.TimeoutError as err:
             # See discussion here: https://github.com/python/asyncio/issues/253#issuecomment-120138132
             # This prevents a nuisance log error message, finished is guaranteed
             # to be cancelled but not awaited when wait_for() has a timeout.
             try:
-                await finished
```
timburke: Actually I think that you need to explicitly assign `outcomes = []` here.  If there was  a timeout then the `finished` future was canceled automatically by `asyncio.wait_for` so when you await it, it will always raise `asyncio.CancelledError` and `outcomes` will never be set. 

I don't think there's a safe way to get any underlying exceptions from the tasks if there was a timeout.

mattrunchey: Ah, yea, good catch, i'll make sure I assign it outside of the area of exception so we don't try and reference it should it never get created.

PR #804, opened 2019-04-29
Fixing incorrect cast to bool

timothyfroehlich:
## Overview

Recently checked in branch was incorrectly converting an boolean argument, some workflows passed the argument as a string that would always be converted to "True" via `bool()`


timothyfroehlich:
@mattrunchey This is the branch I'd like to make the release from. (Heading to lunch righ tnow)

timothyfroehlich:
@mattrunchey I think this should be good.

PR #810, opened 2019-05-01
Add parse_uuid(), parse_uuid2int(), parse_uuid2slug() functions

ekkizogloy:
Add utility functions to aid in the conversions of UUIDs in integer form to slug form and vice versa.

The functions can be called with inputs of many types and they detect and convert into known output types. This is useful when there may be confusion about UUID input types or formats. 

parse_uuid(uuidinput, prefix=None) always returns both the integer form and the slug form. 
parse_uuid2int(uuidinput) always returns just the integer form
parse_uuid2slug(uuidinput, prefix=None) always returns the slug form. 

Input to all the functions can be integers or strings. 
Acceptable string parsing rules are the following:
```
Strings are interpreted as hex if they are prefixed with '0x' including only characters [0-9a-fA-F]
Strings are interpreted as integer if they are not prefixed with '0x' and include only characters [0-9]
Strings can be as a uuid in the form of
                '0000-0000-0000-0000'
                'd--0000-0000-0000-0000'
                'd--0000-0000-0000'
                'd--0000-0000'
                'd--0000'
```

### DIFF
--- a/iotilecore/iotile/core/utilities/gid.py
+++ b/iotilecore/iotile/core/utilities/gid.py
@@ -1,3 +1,4 @@
+import re
 from iotile.core.exceptions import *
 
 # Generate a 0000-0000-0000-0001 given an integer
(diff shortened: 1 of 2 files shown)


iotilecore/iotile/core/utilities/gid.py
```diff
@@ -19,3 +20,89 @@ def uuid_to_slug(uuid):
         raise ArgumentError("Integer should be a positive number and smaller than 0x7fffffff", id=uuid)
 
     return '--'.join(['d', int64gid(uuid)])
+
+
+def parse_uuid(uuidinput, prefix=None):
```
dkarchmer: Please see https://github.com/iotile/python_iotile_cloud/blob/master/iotile_cloud/utils/gid.py#L100

PR #811, opened 2019-05-01
Adding a type option to ModifyJsonStep

timothyfroehlich:
## Overview

Attempting to run this step to set the iotile id would write it out as a string, in quotations.
To alleviate this, this change will allow the type of the value to be set.

## Major Changes

Add additional optional argument to iotile-ship ModifyJsonStep

### DIFF
--- a/iotileship/iotile/ship/actions/ModifyJsonStep.py
+++ b/iotileship/iotile/ship/actions/ModifyJsonStep.py
@@ -48,6 +48,7 @@ class ModifyJsonStep: # pylint: disable=too-few-public-methods
     Args:
         key (list): A list of keys and subkeys specifing the key to change (list can be one)
         value (str): The value to change for the above key
+        value_type (str, optional): The type of value. Typedargs types supported Defaults to 'string'
         path (str): The file path, relative to the Filesystem Manager's root, to the file to change
         create_if_missing (bool, optional): Set to true to create key if the last key in the list is not found
                 Otherwise the function will throw a KeyError. Defaults to False.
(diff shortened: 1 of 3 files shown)


timothyfroehlich:
Can't decide whether the default value should be `string` which is more likely to be picked or `int` which the only current user of ModifyJsonStep uses. 

Or maybe it should just be required.

iotileship/iotile/ship/actions/ModifyJsonStep.py
```diff
@@ -48,6 +48,7 @@ class ModifyJsonStep: # pylint: disable=too-few-public-methods
     Args:
         key (list): A list of keys and subkeys specifing the key to change (list can be one)
         value (str): The value to change for the above key
+        value_type (str, optional): The type of value. Typedargs types supported Defaults to 'string'
```
mattrunchey: I think that, conventionally, the item you put in the parentheses should be referring to the type of the argument. It would be clearer to state that it's an optional argument in the description of the arg.

Also, shouldn't arguments that are optional be grouped at the end (as kwargs)? I know it might not matter in this case since they're all passed under a single "args" input, but in normal functions that is the expected pattern.

mattrunchey:
If production_devices is dependent on this, you'll want to modify the RELEASE.md and version.py, and release it to pypi with this PR too.

iotileship/iotile/ship/actions/ModifyJsonStep.py
timothyfroehlich: Only example I could (easily) find for optional arguments in docstrings suggested that. But I rearranged it.

The value_type being optional directly modifies the value argument so I feel like it should be documented right next to it.

timothyfroehlich:
> If production_devices is dependent on this, you'll want to modify the RELEASE.md and version.py, and release it to pypi with this PR too.

K. Guess I didn't release the last one. How do I do that?

timothyfroehlich:
@mattrunchey For production_devices's module_settings.json what should the python depends line be? My guess is this:

`        "iotile-ship >= 1.0.7",
`

PR #815, opened 2019-05-07
VirtualDevice cleanup step 1

timburke:
## Overview

This PR begins the VirtualDevice cleanup process.  It is into a side branch until we finish collecting the rest of the virtual device fixes.  The first PR includes the following:

1. **Refactor VirtualIOTileDevice:** previously the class had too many bells and whistles that made it annoying to use when you wanted to implement complex behavior because the base class included simple wrappers that obviated the need for complex things sometimes but made implementing those complex things harder.  In particular the fact that you can return streaming and tracing data from a call to `open_interface` and expect it to be traced out of the device was not desirable for all but the simplest test devices.  Similarly most virtual devices don't want thread based periodic background tasks.  

   So, `BaseVirtualDevice` was created that is just the bare minimum required to be a VirtualDevice.  `SimpleVirtualDevice` is a close analog to what `VirtualIOTileDevice` was and a `StandardVirtualDevice` is added that is what most subclasses should inherit from.  

2. **Cleanup Exception Hierarchy** Previously a bunch of rpc related exceptions were incorrectly declared inside `iotile.core.hw.virtual` instead of `iotile.core.hw.exceptions`.  I had previously moved them but there were a number of places within coretools that still imported them from their legacy location.  I've now found and removed all of those imports and removed the legacy exports from `iotile.core.hw.virtual`.

3. **Remove `async_tile_rpc`** This behavior is no longer needed since `tile_rpc` now natively supports coroutine based rpcs.

4. **Refactor `iotile-test` devices** Many of the virtual devices in iotile-test required basic porting to use with the new VirtualDevice hierarchy.  In particular they were moved to inherit from `SimpleVirtualDevice`.

## Side Notes

1. We'll need to update the virtual device tutorial to tell users to inherit from `SimpleVirtualDevice` and they no longer need to provide `@tile_rpc(0x4)` since `SimpleVirtualDevice includes that now.

2. I will need to go through the existing internal emulated devices and make sure they don't use `async_tile_rpc` or have incorrect exception imports.

## Remaining Work

1. Update `VirtualAdapterAsyncChannel` to use coroutines rather than callbacks to allow virtual devices to communicate to attached clients asynchronously.  Since we now have coroutine support throughout `DeviceAdapter` land, we can remove the legacy callback related code.
2. Merge `trace_sync`, `stream_sync` implementations from `EmulatedAdapter` with the new coroutine based versions in `BaseVirtualDevice` that will be implemented in terms of coroutine based functions in `VirtualAdapterAsyncChannel`.
3. Remove old `audit.py` functionality.  This was intended to let you see what rpcs were being sent to a virtualdevice but we now have much better tracing capabilities that obviate the need for this.
4. Reimplement `BaseRunnable` using coroutines rather than worker threads.  
5. Allow passing a loop object into `BaseVirtualDevice` and `BaseVirtualTile` so that they can be placed into a fixed event loop in case they need to spin up coroutines.


iotilecore/iotile/core/hw/virtual/virtualdevice_standard.py
```diff
@@ -0,0 +1,82 @@
...
+            bytes: The response payload from the RPC
+        """
+
+        if rpc_id < 0 or rpc_id > 0xFFFF:
+            raise RPCInvalidIDError("Invalid RPC ID: {}".format(rpc_id))
+
+        if address not in self._rpc_overlays and address not in self._tiles:
+            raise TileNotFoundError("Unknown tile address, no registered handler", address=address)
+
+        overlay = self._rpc_overlays.get(address, None)
```
mattrunchey: I wonder if this function would benefit from having some explanation on what the logic here is doing. When first looking at this function, it may be a bit confusing to see all of these different options for which `call_rpc` method is actually executing somewhere (overlay? tile?). Just a thought. Perhaps this is explained elsewhere already and we could link to it?

iotilegateway/iotilegateway/supervisor/client.py
```diff
@@ -503,18 +503,21 @@ def post_info(self, name, message):
         else:
             try:
                 response = self._rpc_dispatcher.call_rpc(rpc_id, args)
-                if inspect.iscoroutine(response):
+                if inspect.isawaitable(response):
```
mattrunchey: I think I get why this is the better method to use over iscoroutine, but for our purposes does the distinction actually have the possibility of cropping up? Like, I don't see someone implementing `call_rpc` with a generator-based coroutine, right?

iotilegateway/iotilegateway/supervisor/client.py
```diff
@@ -503,18 +503,21 @@ def post_info(self, name, message):
...
                 result = 'invalid_response'
+                response = b''
             except Exception: #pylint:disable=broad-except;We are being called in a background task
                 self._logger.exception("Exception handling RPC 0x%04X", rpc_id)
                 result = 'execution_exception'
+                response = b''
 
         message = dict(response_uuid=tag, result=result, response=response)
-
+        self._logger.warning("Message was %s", message)
```
mattrunchey: Is this meant to be a warning-level log message, or was it left in from debugging?

iotiletest/iotile/mock/devices/realtime_test_device.py
```diff
@@ -27,29 +27,30 @@ class RealtimeTestDevice(VirtualIOTileDevice):
...
-                    in seconds and value is an integer that is sent as the value every interval
-                    as a broadcast reading (BroadcastReport).  The stream id is the key of the
-                    streams dict which should be a string encoding of a hex number including
-                    the prefix 0x so that it can be parsed with int(key, 0).
-
-                trace (list):
-                    A list of tuples which are (float, string) lists
-                    that will trace the fixed string every fixed interval given
-                    by the first float argument in seconds.
+                iotile_id (int) The UUID used for this device.  If no UUID is
```
mattrunchey:
```suggestion
                iotile_id (int): The UUID used for this device.  If no UUID is
```

iotiletest/iotile/mock/devices/realtime_test_device.py
```diff
@@ -27,29 +27,30 @@ class RealtimeTestDevice(VirtualIOTileDevice):
...
+                streams (dict): A map of strings with hex numbers to tuples of
+                    (interval, value) where interval is a float that expresses
+                    how often the stream should stream in seconds and value is
+                    an integer that is sent as the value every interval as a
+                    realtime reading (IndividualReadingReport).  The stream id
+                    is the key of the streams dict which should be a string
+                    encoding of a hex number including the prefix 0x so that
+                    it can be parsed with int(key, 0).
+
+                broadcast (dict): A map of strings with hex numbers to tuples
```
mattrunchey: If you specify multiple broadcasts, I seem to recall the behavior where you would only see one value at a time being broadcast, with the most recent one overwriting whatever was there before. It might help to clarify that behavior here in case someone comes across it?

iotilecore/iotile/core/hw/virtual/virtualdevice_standard.py
timburke: Refactoring to move the `rpc_overlay` functionality out of the base class entirely and into `SimpleVirtualDevice` only, which is the only place that its used.  Now there shouldn't be any more confusion because you **either** get tile based rpc dispatch **or** you get a global rpc dispatch table but never a combination of both.

iotilegateway/iotilegateway/supervisor/client.py
timburke: I'm not sure if we'll see it popping up or not but `isawaitable` is the correct check for what we are trying to discover.  In many cases, `isawaitable` will be identical to `iscoroutine`, which will often be identical to `iscoroutinefunction` called on the parent function but not always and what we actually care about is: `was I handed something that I need to await before I get a result`, which is what `isawaitable directly answers.

An Example where `isawaitable` will work but `iscoroutine` won't is:
- we are returned an `asyncio.Future` object because the RPC itself was dispatched to a background thread pool for long-running execution without blocking the event loop.  In this case we could have forced the RPC implementation to `await` that future, turning the whole function into a coroutine, or it could just return the future and `isawaitable` will tell us we should await it, even though the function that returned it was just a normal function ("pass this object to another thread and get a thing back immediately, return the thing)

iotilegateway/iotilegateway/supervisor/client.py
timburke: left-over, removing.

iotiletest/iotile/mock/devices/realtime_test_device.py
timburke: Fixing in the next push

iotiletest/iotile/mock/devices/realtime_test_device.py
timburke: Added additional comments in the docstring about this

PR #818, opened 2019-05-07
Refactor iotile-emulate to use BackgroundEventLoop

timburke:
## Overview

This is a minimal refactor to make `EmulatedDevice` work with the `SharedLoop` architecture that is now present in `iotile-core`.

Previously EmulatedDevice spun up its own emulation loop in a different thread with its own `asyncio.Loop`.  Now it just adds itself to the chosen `BackgroundEventLoop.`

Note that there is still room for additional refactors to remove some duplicate functionality in EmulationLoop but in the interest of time, it is left for now.  There shouldn't need to be more backwards incompatible changes.

## Major Changes

1. `EmulationLoop` is updated to not build its own `asyncio.Loop`.  It now can be passed a `BackgroundEventLoop` to use and it will use it.  The added features of `EmulationLoop` now are related to handling RPC scheduling and keeping track of which asnycio queues, events are set so that you can see if the device is idle or still chugging through some background work.

2. `BaseRunnable` is moved from using a worker thread per periodic callable to using coroutines running inside of a `BackgroundEventLoop`.

2. Changes 1 and 2 required that VirtualDevices now accept a `BackgroundEventLoop` argument to appropriate run its coroutines in the right event loop so that they can be awaited by `VirtualDeviceAdapter`.  In order to not force all future virtual devices to include a `loop` parameter, VirtualDeviceAdapter now inspects the `__init__` signature of virtual devices and if they take a kwarg named `loop`, it sets it with the correct event loop.  This behavior is documented inside the docstring for `BaseVirtualDevice`.

3. Removed (now duplicate) functionality of `trace_sync` and `stream_sync` that were workarounds to provide coroutine based tracing and streaming apis when VirtualDevice didn't natively support it.  They are now just the same as `trace` and `stream` so they have been removed.

4. Explicitly documented what the `AbstractAsyncDeviceChannel` object is that is passed from VirtualDeviceAdapter to VirtualDevice objects when it starts them and changed the function signatures to be coroutine based rather than callback based since they are now being called in an event loop that can await things.

5. Moved `open_interface` and `close_interface` as well as the interface specific hook functions to support awaiting coroutines.  Previously, only normal functions were supported.

6. Added additional exceptions to have a good error to throw when streaming or tracing data fails.

7. Fixed prior bad behavior where `VirtualDeviceAdapter` would call potentially blocking functions inside of the event loop.  Now VirtualDeviceAdapter is intentional about not running potentially blocking virtual device functions inside the main event loop and support was added for dispatching such things to a background worker thread pool.  Previously we got away with it because `EmulatedDevice` is smart enough to know whether it's being called in a blocking/nonblocking context and do the right thing but since it is now using the same underlying loop as `VirtualDeviceAdapter`, the adapter needs to know when it has to explicitly await things rather than expecting the them to block.

## Notes

1. @mattrunchey This is the last major refactor needed before we cut a release.  I am going to go through and add one more PR with documentation changes based on the new VirtualDevice hierarchy but then it should be good to go.


timburke:
I ended up including the documentation changes in the same PR.

iotilecore/iotile/core/hw/virtual/virtualdevice_simple.py
```diff
@@ -36,13 +37,15 @@ class SimpleVirtualDevice(BaseVirtualDevice, BaseRunnable):
...
         name (str): The 6 byte name that should be returned when anyone asks
             for the controller's name of this IOTile device using an RPC
+        loop (BackgroundEventLoop): The loop we should use for running background
+            tasks. Defaults to the global SharedLoop
     """
 
     __NO_EXTENSION__ = True
 
-    def __init__(self, iotile_id, name):
+    def __init__(self, iotile_id, name, *, loop=SharedLoop):
```
mattrunchey: Should the "*" argument be part of the Args docstring?

timburke: I modified the docstring to indicate that `loop` must be passed as a keyword argument.

timburke:
@mattrunchey FYI I added two more changes here that should make it easier to follow the logic of how rpcs are dispatched:

1. I renamed `BaseVirtualDevice.call_rpc` to `BaseVirtualDevice.async_rpc` and made it always a coroutine since it will always be called from within `VirtualDeviceAdapter.send_rpc` which is itself a coroutine so it may as well always be `async` and avoid the confusion.

2. I ported the `EmulatedDevice.rpc` method which is a blocking debug method that lets you directly send an RPC into a device and wait for the result outside of an event loop.  I renamed it `simple_rpc` and updated the docstring to make it clear that this is for testing/exploration use. It is now a part of `BaseVirtualDevice` as well so that there is a simple way to just instantiate a VirtualDevice class and call one of its RPCs without needing to wrap it inside of a DeviceAdapter.

PR #820, opened 2019-05-08
Merge extended virtual device refactor to master

timburke:
## Overview

This PR moves the last bit of our CoreTools `asyncio` refactor to `master` to prepare for release.


timburke:
> Do we need `version.py` for the three modified packages (iotile-gateway, iotile-core, iotile-emulate) in order to pypi the alpha, or can it be tagged and released anyway with an `rc1` appendage (or should we do it manually?)

Yeah we'll need to bump the versions for a lot of packages.  I'm still figuring out exactly which ones need to be released and bump the versions / release.md for each

PR #826, opened 2019-05-13
Refactor jlink_background.py to use asyncio event loop

DimaWittmann:
## Refactor jlink adapter

- Jlink inherits new StandardDeviceAdapter instead of DeviceAdapter and implements new interface.
- JLinkControlThread renamed to AsyncJlink, removed thread and queue. Now AsyncJlink implement functions to interact with jlink that are called from JlinkAdapter using **run_in_executor** function in order not to block the background thread.
- Add maintenance periodical task for polling control info.

### DIFF
--- a/transport_plugins/jlink/iotile_transport_jlink/jlink.py
+++ b/transport_plugins/jlink/iotile_transport_jlink/jlink.py
@@ -7,10 +7,11 @@ import logging
 import pylink
 from typedargs.exceptions import ArgumentError
 from iotile.core.exceptions import HardwareError
-from iotile.core.hw.transport.adapter import DeviceAdapter
+from iotile.core.hw.transport.adapter import StandardDeviceAdapter
+from iotile.core.utilities import SharedLoop
 from .devices import KNOWN_DEVICES, DEVICE_ALIASES
 from .multiplexers import KNOWN_MULTIPLEX_FUNCS
-from .jlink_background import JLinkControlThread
+from .jlink_background import AsyncJLink
 
 
 # pylint:disable=invalid-name;This is not a constant so its name is okay
@@ -18,7 +19,7 @@ logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
 
-class JLinkAdapter(DeviceAdapter):
+class JLinkAdapter(StandardDeviceAdapter):
     """Wrapper around JLink based transport layer.
 
     Args:
@@ -29,26 +30,27 @@ class JLinkAdapter(DeviceAdapter):
 
     ExpirationTime = 600000
 
-    def __init__(self, port, on_scan=None, on_disconnect=None, **kwargs):
-        super(JLinkAdapter, self).__init__()
+    def __init__(self, port, name=__name__, loop=SharedLoop, **kwargs):
+        super(JLinkAdapter, self).__init__(name, loop)
         self._default_device_info = None
         self._device_info = None
         self._control_info = None
         self._jlink_serial = None
         self._mux_func = None
         self._channel = None
-        self._control_thread = None
+        self._jlink_async = None
         self._connection_id = None
         self.jlink = None
         self.connected = False
+        self.opened_interfaces = {
+            "rpc": False,
+            "tracing": False,
+            "streaming": False,
+            "debug": False,
+            "script": False
+            }
 
         self._parse_port(port)
-
-        if on_scan is not None:
-            self.add_callback('on_scan', on_scan)
-
-        if on_disconnect is not None:
-            self.add_callback('on_disconnect', on_disconnect)
 
     def _parse_port(self, port):
         if port is None or len(port) == 0:
@@ -131,13 +133,17 @@ class JLinkAdapter(DeviceAdapter):
                     print("Warning: multiplexing architecture not selected, channel will not be set")
         return disconnection_required
 
-    def _try_connect(self, connection_string):
+    async def _try_connect(self, connection_string):
         """If the connecton string settings are different, try and connect to an attached device"""
         if self._parse_conn_string(connection_string):
             if self.connected is True:
-                self._trigger_callback('on_disconnect', self.id, self._connection_id)
+                info = {
+                    "reason": "Reconnection",
+                    "expected": True
+                }
+                self.notify_event(connection_string, 'disconnection', info)
                 self.connected = False
-                self.stop_sync()
+                await self.stop()
 
             if self._mux_func is not None:
                 self._mux_func(self._channel)
@@ -147,12 +153,10 @@ class JLinkAdapter(DeviceAdapter):
                                     "or -c device=name in connect_direct or debug command",
                                     known_devices=[x for x in DEVICE_ALIASES.keys()])
 
+            self._jlink_async = AsyncJLink(self, loop=self._loop)
             try:
-                self.jlink = pylink.JLink()
-                self.jlink.open(serial_no=self._jlink_serial)
-                self.jlink.set_tif(pylink.enums.JLinkInterfaces.SWD)
-                self.jlink.connect(self._device_info.jlink_name)
-                self.jlink.set_little_endian()
+                await self._jlink_async.connect_jlink(
+                    self._jlink_serial, self._device_info.jlink_name)
                 self.connected = True
             except pylink.errors.JLinkException as exc:
                 if exc.code == exc.VCC_FAILURE:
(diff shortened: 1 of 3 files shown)


transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
```diff
@@ -1,26 +1,21 @@
 """A background thread for asynchronous operations on the jlink adapter."""
 
+import asyncio
```
timburke: ### Overall Goal

There are two parts of the implementation needed here:

1. Fundamental interaction with `jlink` needs to be wrapped in `coroutines` for consumption by a `StandardDeviceAdapter` subclass whose API is natively coroutine based.  That is the refactor needed in `JLinkControlThread`, which should be renamed `AsyncJLink`.  All of the exposed methods on `AsyncJLink` should be coroutines.
2.  `JLinkDeviceAdapter` needs to restructured so that it subclasses `StandardDeviceAdapter` and implements the `AbstractDeviceAdapter` coroutine API.  It will delegate all of the actual work (connect to jlink, take a jlink action, etc) to `AsyncJLink` to perform and since those methods will be coroutines, the coroutines in StandardDeviceAdapter can nicely `await` them.

This draft PR is targeted at goal 1 above solely so I'll focus comments on that.

### Comments

- we have a common wrapper on top of `asyncio` that all of CoreTools should use: [`BackgroundEventLoop`](https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/utilities/async_tools/event_loop.py#L252).  

- No `asyncio` loops should be created independently in any CoreTools subsysterm.  There should be a single BackgroundEventLoop per process that is declared [here](https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/utilities/async_tools/event_loop.py#L683) (`SharedLoop`), i.e. normally asyncio has a single foreground loop per process, we have a single background loop per process.

- Any long-running blocking style interaction with the jlink hardware (pretty much any call on the `pylink` object) needs to be [run on an executor thread](https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/utilities/async_tools/event_loop.py#L559) so it doesn't block the single shared background event loop.

    You will find many such convenience wrappers around `asyncio` routines that you may choose to use on the `BackgroundEventLoop` class.

- You should not need to manually create any threads in this implementation.  The goal is to move from explicit multithreading to running all background work inside of the `asyncio` loop so we don't achieve that if we are forced to have a mix of threading / asyncio code.

### Suggestions

- In other `asyncio` ports we have found the following structure useful:

   1. Implement the class with an `async` start() and `stop()` coroutine methods.  Defer any blocking work from `__init__` to `start()` so that it can be `awaited`.  This is important since you can't declare `__init__` `async`.  For example, `start()` could connect to `pylink` and setup any necessary structures.  `stop()` could cleanly disconnect from the pylink.

   2. If you need to perform background work that no one is explicitly awaiting, such as continually polling the jlink for events, you can do that by creating a `BackgroundTask` and adding it to the `BackgroundEventLoop`.  This makes sure that it is properly cleaned up when the process exits.  You should create it in `start()` and stop it in `stop()`.  If you have a hierarchy of such tasks, `BackgroundTask` can have children but you shouldn't need that here I don't think.

   3. Always take a `loop=SharedLoop` keyword argument (with default)in `__init__` for any object that interacts with a `BackgroundEventLoop`.  This allows passing a different loop for unit testing.

   4. You should not need to often call `BackgroundEventLoop.get_loop()` to get the raw `asyncio` loop.  Generally there is a better way to achieve your goal using high level `BackgroundEventLoop` methods like `launch_coroutine` or things like `create_event`, `create_queue`, etc.

   5. Never try to have two different asnycio loops running and communicate between them.  It never works well.

DimaWittmann: Thank you for thorough explanation

DimaWittmann: Hi Tim! 
I wanted to ask you about the point:
_Any long-running blocking style interaction with the jlink hardware (pretty much any call on the pylink object) needs to be run on an executor thread so it doesn't block the single shared background event loop._
There are some functions, **async def send_rpc** for example, in AsyncJlink that contains a lot of jlink IO calls. Is this mandatory for AsyncJlink to provide API using coroutines only or it would be okay if **def send_rpc** will be called in ** run_in_executor ** as a whole?

timburke: I don't think it's a hard requirement what the interface is between `AsyncJLink` and `JLinkDeviceAdapter`, though in this case I think it would actually be easier to make `send_rpc` async for the following reasons:

- `send_rpc` now is really just a `write_memory` followed by a `trigger_interrupt` and then a polling loop.  `write_memory` will already need to be `async` to call from `async debug`.  `trigger_interrupt` is easy to make async, so then it comes down to the polling loop.

   That loop is going to need to go away most likely anyway since the jlink needs to constantly be checking for new tracing or streaming data out of the device and those checks can be merged together with the checks for whether an RPC is finished into a single background asyncio task.

   So, given the need to support tracing and streaming, `send_rpc` should become a simple function whose core building block `write_memory` already got an asyncio wrapper for use elsewhere.

That being said, there is no hard requirement on what the `AsyncJLink` API is beyond that:
- there are no background threads spun up for anything (use executor calls instead if needed)
- the main asyncio loop (`SharedLoop`) is never blocked
- it works correctly

DimaWittmann: I am just trying to understand if I am wrong. 
_send_rpc_ is just a bunch of _jlink.read_mem_ and _jlink.write_mem_ functions. These functions are considered blocking. To me it looks like just calling _send_rpc_ and other help functions in executor would resolve possible problems of blocking BackgroundLoop as well as it shows clearly intention (interaction with jlink should be async and in separate thread).
Do I miss something?   

I updated PR so I could see what I mean, I will appreciate any comments from your side.

timburke: I understand the idea and, of course it would work, I just think that you would have a lot of places in your code with a `loop.run_in_executor(self.read_mem, ...)` whenever you wanted to read_mem.  

If you instead did something like (rename `read_mem` to `_read_mem_blocking` and then did:
```
async def read_mem():
    loop.run_in_executor(self._read_mem_blocking, ...)
```

It would be clear that `read_mem` is an async function (it's marked as such) whereas the blocking variety is also clearly marked (and internal so no one calls it by mistake.  Then you could build up an async send_rpc as just:

```
async def send_rpc():
    await async_jlink.write_mem()
    await async_jlink.trigger_interrupt()

    await async_jlink.rpc_complete_event()
```

Where `rpc_complete_event` is an `asyncio.Event()` that is being set by the background asyncio reader task that is polling for rpc completion and checking for streaming/tracing data.

DimaWittmann: I see your point. Thank you for explanations. I updated as you suggest, still have to refactor and add handling events for rpc and streaming/tracing

DimaWittmann: I was wondering if it's better for you to review separate PR refactoring + tracing/streaming or everything in the single PR? I made all changes regarding refactoring I wanted. Please, feel free to share you thoughts.

timburke: Generally I would prefer smaller PRs with each feature as you did above to make it easier and quicker to review rather than a giant change set.  

I'm taking a look now at the PR and will give feedback.

transport_plugins/jlink/iotile_transport_jlink/jlink.py
```diff
@@ -163,166 +160,82 @@ def _try_connect(self, connection_string):
...
+        """Asynchronously stop this adapter and release all resources."""
 
-        if self._control_thread is not None and self._control_thread.is_alive():
-            self._control_thread.stop()
-            self._control_thread.join()
+        await self._jlink_async.close_jlink()
 
-        if self.jlink is not None:
-            self.jlink.close()
+    async def probe(self):
```
timburke: The `probe` API function can be called without a connection established and should not fail even if the connected device has blank flash or is hard-faulted, etc.  The usage in CoreTools is:

- For device adapters that don't gratuitously send `device_seen` events, they set a probe_required flag saying they need to polled to see if a device is present. 
- whenever a user calls `scan` on `HardwareManager` and the device adapter requires probing, it will call `probe`.  `probe` should ensure that any connected devices generate a `device_seen` event before returning (as you do below)
- if there are no connected devices that have a uuid then probe returns without error (since the probe worked) but no device_seen are generated.

So if there is no control info block found, probe shouldn't fail, it should just not generate any `device_seen` events.  Also, probe cannot assume anything about the status of a connection.  It may be called both during a connection and when there is no connection.  If there is a connection and that somehow means that it cannot probe for a control structure (which I don't believe is the case) then it can generate a `device_seen` event based on cached data for the current connection.

transport_plugins/jlink/iotile_transport_jlink/jlink.py
```diff
@@ -163,166 +160,82 @@ def _try_connect(self, connection_string):
...
+        self._setup_connection(conn_id, connection_string)
 
-        def _on_finished(_name, control_info, exception):
-            if exception is not None:
-                callback(connection_id, self.id, False, str(exception))
-                return
+        try:
+            self._control_info = await self._jlink_async.verify_control_structure(self._device_info, self._control_info)
+        except HardwareError as e:
+            logger.exception("RPC, streaming and tracing won't work, but device still can be flashed or debugged")
```
timburke: I would move this down to `warning` level and add `exc_info=True` to still attach the exception information to the event.  It's technically expected that this could happen in some situations so it's not a proper error,

transport_plugins/jlink/iotile_transport_jlink/jlink.py
```diff
@@ -163,166 +160,82 @@ def _try_connect(self, connection_string):
...
 
-            if control_info is not None:
-                self._control_info = control_info
+        self._connection_id = conn_id
 
-            callback(connection_id, self.id, True, None)
-            self._connection_id = connection_id
+    async def open_interface(self, conn_id, interface):
```
timburke: I don't believe this will work as implemented.  There are multiple interfaces that can be opened and closed independently:

- `rpc`, `tracing`, `streaming`, `debug`, `script`

`debug` and `scrip` open/close may be no-ops for jlink and `rpc`, `tracing`, `streaming` may have the same underlying `poll` method feeding data to each, however:
- `trace` events must only be generated when `tracing` interface is open
- `report` events must only be generated when `report` interface is open

Also, the status of the `trace` and `streaming`  interfaces needs to be communicated to the device so that it can implement flow control.  For example, inside the controller the `stream_manager` subsystem needs to know whether the `streaming` interface is open/closed because it calls the `open_stream` rpc and that must fail if no jlink adapter is connected and no one is listening for tracing/report data.

So, it seems like you need two different things here:

1. If any of `rpc`, `tracing`, `streaming`  are open, then `polling` is on
2. A separate flag for whether each individual interface is opened.  The `tracing` and `streaming` interface flags need to be communicated to the firmware so that it knows if it can send tracing/streaming data.  

Additionally, stale data needs to be cleared (if present) on opening `tracing`/`streaming` interfaces if it's possible for stale data to be stuck in the fifo.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
```diff
@@ -1,188 +1,113 @@
...
+        self._jlink = None
+        self._loop = loop
 
-    def stop(self):
-        """Tell this thread to stop, do not wait for it to finish."""
-        self._commands.put(JLinkCommand(JLinkControlThread.STOP, None, None))
+        # maitenance can be enabled for different interfaces
+        self._maitenance_counter = 0
+        self._maitenance_task = None
+        self.rpc_response_event = asyncio.Event() # triggered when the controller acknowledge rpc response
```
timburke: All `asyncio` primitives like `Event`, `Lock`, etc need to be created by requesting from the `BackgroundEventLoop` passed in.  If you request directly from `asyncio` it will create a different `asyncio` loop and attach the event to that.

You can see this if you run `iotile` tool with `-vvvvv` option and look at the logs.  You should only ever see **one line** that says something like "Starting asyncio event loop using method SelectorSomething".  If you see multiple lines like that it means that you're asyncio primitives are not all attached to the same loop and the code **will break** in subtle and hard to diagnose ways.

The safe way to get things from `asyncio` is to request them through the `loop` object you were passed in to ensure you don't accidentally create another loop.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
```diff
@@ -262,76 +188,142 @@ def _debug_read_memory(self, device_info, _control_info, args, _progress_callbac
...
+            self._connect_jlink_blocking, jlink_serial, jlink_name)
+
+        return self._jlink
 
-    def _read_memory(self, start_address, length, chunk_size=4, join=True):
+    async def close_jlink(self):
+        if self._jlink is not None:
+            await self._loop.run_in_executor(self._jlink.close)
+
+    async def start_polling(self, control_info, step_timeout=0.1):
```
timburke: I see here the counting approach to handing polling.  That works and you can ignore the previous comment I made about needing to do that, I hadn't read this far in the code yet.  I still think you need to individually track the status of the various interfaces but the overall polling approach makes sense to me now.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
```diff
@@ -262,76 +188,142 @@ def _debug_read_memory(self, device_info, _control_info, args, _progress_callbac
...
+        counter = self._maitenance_counter
```
timburke: `maintenance` is consistently misspelled

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
```diff
@@ -262,76 +188,142 @@ def _debug_read_memory(self, device_info, _control_info, args, _progress_callbac
...
 
-    def _debug_write_memory(self, _device_info, _control_info, args, _progress_callback):
+    async def debug_write_memory(self, _device_info, _control_info, args):
         memory_region = args.get('region').lower()
 
         if memory_region == 'external' or memory_region == 'flash':
             raise ArgumentError("Writing to flash/external is not currently supported.")
-        self._jlink.memory_write(args.get('address'), args.get('data'))
+        await self._loop.run_in_executor(
+            self._jlink.memory_write, args.get('address'), args.get('data'))
```
timburke: Can't this just use the `async write_memory` method that you define below rather than running in an executor thread?

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
DimaWittmann: May you please create *feat-jlink-upgrade* branch. I am not sure I have access rights to do it.

transport_plugins/jlink/iotile_transport_jlink/jlink.py
DimaWittmann: Thank you for explanation. Basically probe either reports a connected device or does nothing, am I right?

transport_plugins/jlink/iotile_transport_jlink/jlink.py
DimaWittmann: Will do so

transport_plugins/jlink/iotile_transport_jlink/jlink.py
DimaWittmann: I will add flags as you suggest

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
DimaWittmann: It's my bad, forget about background loop.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
timburke: Done, branch exists now

transport_plugins/jlink/iotile_transport_jlink/jlink.py
timburke: Yes, except I want to be clear on the meaning of `connected`.  As a general point `probe` reports a device that **is available for connection**, i.e. calling `connect <UUID>` with the UUID returned by probe should be expected to work (assuming you're not already connected), or does nothing.  

For the jlink that means the device is:
- physically connected to the jlink
- has functioning firmware with a control structure in RAM to process RPCs

DimaWittmann: I see, thank you for the answer, I updated probe so that now it tries to connected to any of KNOWNDEVICES and find control structure in RAM.

PR #829, opened 2019-05-21
Fix cloud uploader to support very old controller firmware.

timburke:
Hotfix to support uploading readings from very old pods that don't support configuring the maximum report size.  This is a customer issue currently being hit in Zambia on some very old POD-1 family devices with out-of-date firmware.

The cloud uploader now supports those firmwares by making report size configuration opportunistic and not failing if the RPC is not supported.

@ekkizogloy Note that the default maximum report size on those PODs is 10k readings so if the customer has more than 10k readings they will need to do multiple uploads (making sure the data is acknowledged / processed by the cloud between each upload) in order to get all of the data off the POD.

### DIFF
--- a/iotile_ext_cloud/RELEASE.md
+++ b/iotile_ext_cloud/RELEASE.md
@@ -2,6 +2,11 @@
 
 All major changes in each released version of the iotile-ext-cloud plugin are listed here.
 
+## 1.0.7
+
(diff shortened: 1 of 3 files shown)


timburke:
@mattrunchey Per our new release guidelines, I believe this qualifies for immediate patch release since it's currently affecting a customer in the field.

PR #832, opened 2019-05-29
Jlink tracing and streaming

DimaWittmann:
Add handling trace and stream flags over jlink.
Add polling debug queue for traces and reports.
Add handling watch counter.
Should be merged after https://github.com/iotile/coretools/pull/826

### DIFF
--- a/iotiletest/iotile/fixture/device_fixture.py
+++ b/iotiletest/iotile/fixture/device_fixture.py
@@ -60,7 +60,7 @@ def device(port, device_id, direct, record, request):
             except HardwareError:
                 time.sleep(1)
 
-        if i == max_attempts:
+        if i == max_attempts - 1:
             pytest.fail("Could not connect to device after %d attempts, failing" % max_attempts)
             return
 
@@ -92,7 +92,7 @@ def per_test_device(port, device_id, direct, record, request):
             except HardwareError:
                 time.sleep(1)
 
-        if i == max_attempts:
+        if i == max_attempts - 1:
             pytest.fail("Could not connect to device after %d attempts, failing" % max_attempts)
             return
 
--- a/transport_plugins/jlink/iotile_transport_jlink/jlink.py
+++ b/transport_plugins/jlink/iotile_transport_jlink/jlink.py
@@ -7,7 +7,9 @@ import logging
 import pylink
 from typedargs.exceptions import ArgumentError
 from iotile.core.exceptions import HardwareError
+from iotile.core.hw.exceptions import DeviceAdapterError
 from iotile.core.hw.transport.adapter import StandardDeviceAdapter
+from iotile.core.hw.reports import IOTileReportParser
 from iotile.core.utilities import SharedLoop
 from .devices import KNOWN_DEVICES, DEVICE_ALIASES
 from .multiplexers import KNOWN_MULTIPLEX_FUNCS
@@ -29,6 +31,7 @@ class JLinkAdapter(StandardDeviceAdapter):
     """
 
     ExpirationTime = 600000
+    POLLING_INTERFACES = ["rpc", "tracing", "streaming"]
 
     def __init__(self, port, name=__name__, loop=SharedLoop, **kwargs):
         super(JLinkAdapter, self).__init__(name, loop)
@@ -49,6 +52,8 @@ class JLinkAdapter(StandardDeviceAdapter):
             "debug": False,
             "script": False
             }
+        self.report_parser = IOTileReportParser(
+            report_callback=self._on_report, error_callback=self._on_report_error)
 
         self.set_config('probe_required', True)
         self.set_config('probe_supported', True)
@@ -135,6 +140,20 @@ class JLinkAdapter(StandardDeviceAdapter):
                 else:
                     print("Warning: multiplexing architecture not selected, channel will not be set")
         return disconnection_required
+
+    def _on_report(self, report, context):
+        conn_string = self._get_property(self._connection_id, 'connection_string')
+        self.notify_event(conn_string, 'report', report)
+
+        return False
+
+    def _on_report_error(self, code, message, context):
+        print("Report Error, code=%d, message=%s" % (code, message))
+        self._logger.critical("Error receiving reports, no more reports will be processed on this adapter, code=%d, msg=%s", code, message)
+
+    def add_trace(self, trace):
+        conn_string = self._get_property(self._connection_id, 'connection_string')
+        self.notify_event(conn_string, 'trace', trace)
 
     async def _try_connect(self, connection_string):
         """If the connecton string settings are different, try and connect to an attached device"""
@@ -271,7 +290,11 @@ class JLinkAdapter(StandardDeviceAdapter):
         self._setup_connection(conn_id, connection_string)
 
         try:
-            self._control_info = await self._jlink_async.verify_control_structure(self._device_info, self._control_info)
+            self._control_info = await self._jlink_async.verify_control_structure(
+                self._device_info, self._control_info)
+
+            await self._jlink_async.change_state_flag(
+                self._control_info, AsyncJLink.CONNECTION_BIT, True)
         except HardwareError as e:
             logger.warning("RPC, streaming and tracing won't work, but device still can be flashed or debugged",
                 exc_info=True)
@@ -279,8 +302,20 @@ class JLinkAdapter(StandardDeviceAdapter):
         self._connection_id = conn_id
 
     async def disconnect(self, conn_id):
-        await self._jlink_async.close_jlink()
         self._teardown_connection(conn_id)
+
+        if not self._control_info: # direct debug connection case
+            return
+
+        try:
+            for interface in JLinkAdapter.POLLING_INTERFACES:
+                if self.opened_interfaces[interface]:
+                    await self.close_interface(conn_id, interface)
+
+            await self._jlink_async.change_state_flag(
+                self._control_info, AsyncJLink.CONNECTION_BIT, False)
+        except pylink.errors.JLinkReadException:
+            logger.warning("Error disconnecting jlink adapter", exc_info=True)
 
     async def open_interface(self, conn_id, interface):
         self._ensure_connection(conn_id, True)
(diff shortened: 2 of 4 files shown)


timburke:
@DimaWittmann I have merged the other PR into `feat-jlink-upgrade` so this one can be rebased now.

DimaWittmann:
@timburke I added to my changes tracing and streaming support, you could find it in this PR as well as in two others:
https://github.com/iotile/lib_controller/pull/181/files
https://github.com/iotile/con_btc1/pull/159

At the moment I am checking behavior with big reports and also test suite. The issue I see know is on add_streamer command if there are multiple report available not all of them arrives as streamers gets corrupted and stops responding to new data or trigger command.
 Please, share your suggestions regarding implementation and possible cases you would like me to check additionally to those provided by colleagues from Arch. 

Also I would like to ask you about handling `reset` rpc. As it should reset the controller it makes no sense to poll for response. Do you have any thought about this rpc?
Thank you,
Dmytrii

iotiletest/iotile/fixture/device_fixture.py
```diff
@@ -50,7 +50,7 @@ def device(port, device_id, direct, record, request):
         # before giving up to improve test robustness.
         max_attempts = 3
         i = 0
-        for i in range(0, max_attempts):
+        for i in range(0, max_attempts + 1):
```
timburke: Curious about the logic here.  This change should make it try `max_attempts + 1` times instead of `max_attempts` times.  If `3` attempts isn't enough `max_attempts` should just be increased.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
```diff
@@ -40,7 +47,13 @@ def __init__(self, jlink_adapter, loop=SharedLoop):
         await self.write_memory(write_address, write_data, chunk_size=4)
 
         await self._trigger_rpc(device_info)
-        await self._wait_rpc_completion(control_info, timeout)
+
+        if rpc_id == self.RESET_RPC_ID:
```
timburke: This also needs to check if `address == 8` since you can send this RPC to a peripheral tile which will not cause the `jlink` to disconnect because it just resets the tile's mcu.

transport_plugins/jlink/iotile_transport_jlink/jlink.py
```diff
@@ -271,32 +289,70 @@ def _parse_conn_string(self, conn_string):
...
+                self._control_info, AsyncJLink.CONNECTION_BIT, False)
+        except pylink.errors.JLinkReadException:
+            logger.warning("Error disconnecting jlink adapter", exc_info=True)
+
         self._teardown_connection(conn_id)
 
     async def open_interface(self, conn_id, interface):
         self._ensure_connection(conn_id, True)
 
+        if not self._control_info:
```
timburke: This error chomping should not be necessary.  If there is no control_info and someone tries to open an interface, the attempt should fail because that interface could not be opened.  The bug is with the calling code not this code.  @DimaWittmann Where are you seeing errors that this code is working around because we need to fix those instead.

transport_plugins/jlink/iotile_transport_jlink/jlink.py
```diff
@@ -271,32 +289,70 @@ def _parse_conn_string(self, conn_string):
...
+                self._control_info, AsyncJLink.STREAM_BIT, True)
+
+        if interface in JLinkAdapter.POLLING_INTERFACES:
             await self._jlink_async.start_polling(self._control_info)
 
     async def close_interface(self, conn_id, interface):
         self._ensure_connection(conn_id, True)
 
-        if interface in ["rpc", "tracing", "streaming"]:
+        if not self._control_info:
```
timburke: Same comment as above, this error chomping should not happen here and instead we need to fix the code that is calling `close_interface` gratuitously

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
```diff
@@ -358,19 +384,92 @@ def _read_memory_blocking(self, start_address, length, chunk_size=4, join=True):
...
+
     async def _maintenance_coroutine(self, control_info, step_timeout):
         while self._maintenance_counter > 0:
-            if self._jlink_adapter.opened_interfaces["rpc"]:
+            try:
                 await self._poll_rpc_status(control_info)
+                while not await self._poll_queue_status(control_info):
+                    pass
+                await self._update_watch_counter(control_info)
+            except concurrent.futures.CancelledError:
```
timburke: This should not be the exception that you receive here.  `concurrent.futures.CancelledError` is thrown only by multithreading code.  Since we're using asyncio tasks, the exception will be `asyncio.CancelledError`.

iotiletest/iotile/fixture/device_fixture.py
DimaWittmann: The thing is that below, there is a check i == max_attempts, that is not possible if a exclusive range is used

timburke: Ah, thanks for the clarification.  Can you just fix the below line to be `max_attempts - 1`?

transport_plugins/jlink/iotile_transport_jlink/jlink.py
timburke: Note as a caveat to this that the `debug` interface does not require control_info (and opening it is a no-op).  For all other interfaces, opening/closing them should fail if there is no config info

DimaWittmann: I see your point, I will review these checks

iotiletest/iotile/fixture/device_fixture.py
DimaWittmann: Oh course

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
DimaWittmann: Thank you

transport_plugins/jlink/iotile_transport_jlink/jlink.py
DimaWittmann: Yes, you are right

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
DimaWittmann: Fixed it.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
```diff
@@ -33,14 +40,22 @@ def __init__(self, jlink_adapter, loop=SharedLoop):
...
         await self.write_memory(write_address, write_data, chunk_size=4)
 
         await self._trigger_rpc(device_info)
-        await self._wait_rpc_completion(control_info, timeout)
+
+        if rpc_id == self.RESET_RPC_ID and address == self.CONTROLLER_ADDRESS:
+            await asyncio.sleep(2)
+            await self._update_state_flags_after_reset(device_info, control_info)
+            return None
+        elif best_effort:
```
timburke: It seems like this is a race condition: if an rpc with best effort is called followed by a normal rpc, the next rpc can be dispatched before the previous rpc finishes.  Given the intent, I believe, is to allow `notify_sensorgraph` to not hang `close_interface` if the device connection is lost, here's an alternative:

- remove `best_effort`
- wrap call to `notify_sensorgraph` in try/except to catch the timeout failure if the rpc fails.
- log that failure so that `notify_sensorgraph succeeds.

This removes the race and the need for special-casing inside `send_rpc`.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
```diff
@@ -33,14 +40,22 @@ def __init__(self, jlink_adapter, loop=SharedLoop):
...
         write_address, write_data = control_info.format_rpc(address, rpc_id, payload)
         await self.write_memory(write_address, write_data, chunk_size=4)
 
         await self._trigger_rpc(device_info)
-        await self._wait_rpc_completion(control_info, timeout)
+
+        if rpc_id == self.RESET_RPC_ID and address == self.CONTROLLER_ADDRESS:
+            await asyncio.sleep(2)
+            await self._update_state_flags_after_reset(device_info, control_info)
+            return None
```
timburke: In keeping with the other device adapters, this should raise `TileNotFound` instead of returning None.  The rest of coretools handles this exception correctly when a reset RPC is sent.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
```diff
@@ -33,14 +40,22 @@ def __init__(self, jlink_adapter, loop=SharedLoop):
...
+        elif best_effort:
+            return None
+        else:
+            await self._wait_rpc_completion(control_info, timeout)
```
timburke: I can't comment on the actual line since it's not in this diff but:

- `_wait_rpc_completion` should raise `DeviceAdapterError` not `HardwareError` for timeout and communication issues.

`None` is not a valid return value from `send_rpc`.  Either it should return the `bytes` result of the RPC call or raise an exception because the RPC call did not succeed.

transport_plugins/jlink/iotile_transport_jlink/jlink.py
```diff
@@ -318,7 +372,7 @@ def _parse_conn_string(self, conn_string):
         response = await self._jlink_async.send_rpc(
             self._device_info, self._control_info,
             address, rpc_id, payload, timeout)
-        return response['payload']
+        return response['payload'] if response else None
```
timburke: See comment below, either `bytes` should be returned (possibly 0 length) or an exception should be raised.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
DimaWittmann: You are right, I wasn't sure how to handle `close_interface` on reset, thank you for advise.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
DimaWittmann: Thank you for explanation, fixed it.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
DimaWittmann: I see what you mean, should I replace other `HardwareError`s with `DeviceAdapterError`?

transport_plugins/jlink/iotile_transport_jlink/jlink.py
DimaWittmann: Fixed it

DimaWittmann:
>     * `logistics_tracker` download works
>     * `pytest` controller unit tests pass
>     * `iotile hw debug -c <connection string> flash` is able to flash an image onto a bare mcu without working formware
>     * `iotile hw debug flash` works on both nrf52832 and lpc824 targets

@timburke Regarding testing at my side: 

* I was able to start/stop/view trip results but I can't upload logistic_tracker data as I don't have credentials.
* `lib_controller` tests passed except two of it.
* I was able to flash both controller on nrf52832 and firm_accel on tile 2 via jlink.

transport_plugins/jlink/iotile_transport_jlink/jlink_background.py
timburke: Yes but that can be done later.  Let's merge this one and then setup a final round of testing here.

PR #835, opened 2019-05-30
Fix bug in utc assigner which resulted in oscillating time assignments

ekkizogloy:
This fixes a bug in the utc_assigner where adjacent timestamps would oscillate between two values. This occurred because the code path in the 'fix_left()' assignment function would skip the left adjacent data point. If there was a large difference between the two, then the resulting output would oscillate. 

The zip file attached contains a set of problematic report *bin files as well as a script to assign timestamps (fix.py) and the resulting text file versions of the inputs and outputs. 

[pr.zip](https://github.com/iotile/coretools/files/3237674/pr.zip)


slack0:
@ekkizogloy When the fix is ready, is it possible to update any past devices' data with the fix - so that reports could be regenerated? I can verify generating the analytics reports with the fix.

mattrunchey:
One thing to note is that we've recently updated the releasing guidelines that cover the type of changes that justify a patch release -- that is, we want to try defaulting to HEAD if possible and batch minor changes less frequently. https://github.com/iotile/coretools/blob/master/RELEASING.md#patch-change-rules

Is this change immediately needed for an external user/customer, or can we try out the new process? The change will be available in master for dev installs regardless.

iotilecore/test/test_reports/test_utc_assigner.py
```diff
@@ -166,3 +168,69 @@ def test_whole_report_fixing(assigner):
...
+    assigner.add_report(report1)
+    assigner.add_report(report2)
+
+    fixed0 = assigner.fix_report(report0)
+    fixed1 = assigner.fix_report(report1)
+    fixed2 = assigner.fix_report(report2)
+
+    time.sleep(1)
+
+    #print("\nReport 0")
```
mattrunchey: Should clean up this comment block before merging

iotilecore/test/test_reports/test_utc_assigner.py
```diff
@@ -166,3 +168,69 @@ def test_whole_report_fixing(assigner):
...
+    assigner2.add_report(report2)
+
+    fixed0_2 = assigner2.fix_report(report0)
+    fixed1_2 = assigner2.fix_report(report1)
+    fixed2_2 = assigner2.fix_report(report2)
+
+    compare_fixed_report(fixed0_2, 'd_05db/report_0_05db_fixed.txt')
+    compare_fixed_report(fixed1_2, 'd_05db/report_1_05db_fixed.txt')
+    compare_fixed_report(fixed2_2, 'd_05db/report_2_05db_fixed.txt')
+
```
mattrunchey: Extra lines at end of file (PEP8 - should only be one)

iotilecore/test/test_reports/test_utc_assigner.py
```diff
@@ -63,6 +64,8 @@ def compare_fixed_report(fixed, ref_filename):
     assert len(fixed.visible_readings) == len(ref_lines)
 
     for reading, ref_string in zip(fixed.visible_readings, ref_lines):
+        if str(reading) != ref_string:
+            print("FAIL: {} != {}".format(str(reading),ref_string))
```
mattrunchey: Is this check redundant with the `assert`? The assert below should take care of this check now.

iotilecore/version.py
```diff
@@ -1 +1 @@
-version = "5.0.0"
+version = "5.0.1"
```
mattrunchey: Marking a comment in the review to make sure we answer the question justifying a patch release now, so that we can have practice with the process.

iotilecore/test/test_reports/test_utc_assigner.py
```diff
@@ -166,3 +168,69 @@ def test_whole_report_fixing(assigner):
...
+
+    assigner.add_report(report0)
+    assigner.add_report(report1)
+    assigner.add_report(report2)
+
+    fixed0 = assigner.fix_report(report0)
+    fixed1 = assigner.fix_report(report1)
+    fixed2 = assigner.fix_report(report2)
+
+    time.sleep(1)
```
timburke: I believe this and the `import time` line at the top should be removed.  It just slows down the test and should not be necessary.

iotilecore/version.py
timburke: Thanks for flagging the process here Matt.  In this case I believe immediate release is justified because the bug is hit by an external customer and it's impact is particularly bad.  It results in all data uploaded to the cloud being wrong...

iotilecore/test/test_reports/test_utc_assigner.py
mattrunchey: Sounds good. Definitely a fair justification and now we have it recorded here for posterity :)

iotilecore/version.py
mattrunchey: Sounds good. Definitely a fair justification and now we have it recorded here for posterity :)

mattrunchey:
Note justifying immediate release from @timburke :

> Thanks for flagging the process here Matt. In this case I believe immediate release is justified because the bug is hit by an external customer and it's impact is particularly bad. It results in all data uploaded to the cloud being wrong.

ekkizogloy:
Everything should be cleaned up now. Let me know what you think. Tests added to check for monotonicity and repeatability. Both fail without fix.

iotilecore/test/test_reports/test_utc_assigner.py
mattrunchey: @ekkizogloy I think you missed this one

mattrunchey:
@ekkizogloy just noticed that one of the checked in files is a full 48k lines. Is that entire file used for the test or is it shrinkable?  It stood out to me

ekkizogloy:
The test data added is the full data file for a POD. It can't really be shrunk without changing the report/test data. I'd like to keep this test as is because it represents an actual problematic case.

PR #845, opened 2019-06-27
Augment TileBasedDevice to have some virtual controller functionality

mattrunchey:
The goal of this feature is to enable products to use tie tile_manager functionality of a controller
without actually having a controller (virtual devices).

## Overview

These changes cover one areas:
- Responding to tile_manager requests for tile details on each slot

## Major Changes

There is a new VirtualController class in the tile_based_device.py file that is included in all TileBasedDevices. By default you do not need to complete anything and placeholders will be used. The tile_manager will function always, however.

### Reason for version bump

This virtual controller has an intended use case in the next release of one of our products soon and we need a released version of coretools to install it. It introduces new endpoints but doesn't modify existing ones, which warrants a patch bump.

### DIFF
--- a/iotilecore/iotile/core/hw/virtual/tile_based_device.py
+++ b/iotilecore/iotile/core/hw/virtual/tile_based_device.py
@@ -2,11 +2,10 @@
 
 The device handles RPCs by dispatching them to tiles configured using a config dictionary.
 """
-
-from iotile.core.exceptions import ArgumentError
-from ..exceptions import RPCInvalidIDError, RPCNotFoundError, TileNotFoundError
+from iotile.core.hw.virtual import tile_rpc
 from .virtualdevice_standard import StandardVirtualDevice
 from .virtualtile import VirtualTile
+
 
 class TileBasedVirtualDevice(StandardVirtualDevice):
     """A VirtualDevice composed of one or more tiles.
@@ -14,10 +13,14 @@ class TileBasedVirtualDevice(StandardVirtualDevice):
     Args:
         args (dict): A dictionary that lists the tiles that
             should be loaded to create this virtual device.
+        override_controller (bool): Controls whether the virtual controller
+            gets added or not. Defaults to false.
     """
 
-    def __init__(self, args):
+    def __init__(self, args, override_controller=False):
         iotile_id = args.get('iotile_id')
+
+        con_args = args
 
         if isinstance(iotile_id, str):
             iotile_id = int(iotile_id, 16)
@@ -34,6 +37,13 @@ class TileBasedVirtualDevice(StandardVirtualDevice):
             tile = tile_type(address, args, device=self)
 
             self.add_tile(address, tile)
+
+        # Add a controller tile
+        if not override_controller:
+            tile = VirtualController(8, con_args, device=self)
+            self.add_tile(8, tile)
+            tile.tiles = self._tiles
+            tile.prep_count()
 
     def start(self, channel):
         """Start running this virtual device including any necessary worker threads.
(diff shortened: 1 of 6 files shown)


mattrunchey:
@timburke - In particular, I modified the parser and initialization of virtual devices to contain a new parameter for whether to persist the config file changes, as well as the location of the config file (since it wasn't previously accessible in the tile_based_device before). If there is a different way that I should implement this, let me know. Also, I duplicated the code for unpack and pack from iotile_emulate. Do you think we should move those functions to a more common utility location or is that overkill for this? 

@timothyfroehlich -- The RPC definitions here are intended to be used by the ota_updater process we are working on elsewhere. Essentially, that process will need to connect to the virtual_device to query for current version information, as well as to update that version when the process is committed. I would expect calls to look a lot like the ones I implemented in the `test_vircon` function (in `test_virtualadapter.py`). Also take note of the `tile_config.json` -- that is where I added the tags (this is intended to be equivalent location wise to our access_point_config.json and other node_config.json for the future)

mattrunchey:
@timburke @timothyfroehlich ping for review (notes about particulars in above comment)

iotilecore/iotile/core/hw/virtual/tile_based_device.py
```diff
@@ -55,3 +69,207 @@ def stop(self):
...
+    @tile_rpc(0x1008, "", "L8xLL")
+    def controller_info(self):
+        """Get the controller UUID, app tag and os tag."""
+        return [self.config.get('iotile_id'),
+                _pack_version(*(self.config.get('os_tag', 0), self.config.get('os_version', "0.0"))),
+                _pack_version(*(self.config.get('app_tag', 0), self.config.get('app_version', "0.0")))]
+
+    @tile_rpc(0x100B, "LLBB", "L")
+    def update_os_app_tag(self, packed_os, packed_app, set_os, set_app):
+        """Update the OS and/or APP tag of the device.
```
timburke: @mattrunchey I'm not sure I agree that we need to implement this functionality here for a couple of reasons:

- The RPC existed on PODs because there was no other way to set the os/app tag in manufacturing but for virtual devices, it's straightforward to just edit the json config file directly if needed.
- The rest of the `VirtualTile` infrastructure doesn't make any assumption that it will be reading a config from a json file or try to update it in any way so we're locking something in that preclude getting configuration from an in-memory dict or other mechanism.  

Can you explain the use-case here for why we want to add this?  I would lean toward just leaving it out and implementing the rest of the RPCs you have here absent a clear reason why we need it.

mattrunchey: For keeping track of node versions that was persistent, we needed some place on the file system to store the version information. Rather than implementing it outside of the virtual tile (i.e. in a task), I noticed that these RPCs existed to do that same thing for PODs. We need these RPCs somewhere, to update version information, and we decided on using a config json (if that's a bad idea, that's a separate discussion). Since version for PODs was taken care of in the controller, I figured that our VirtualDevices might benefit from being able to do the same.

So the alternative to this would be to move all of the version information to the task subsystem that we use for OTA and not add it to the core VirtualDevice side of the picture. It would be the same RPCs, just tied to the lifecycle and configuration attributes of a task compared to the controller. If other actors wanted to use it (i.e. the factory companion app), they would need to determine if the node they're talking to has an OTA subsystem, and use the RPCs from that tile instead.

I chose to put it in the virtualized controller rather than the task because it seemed that tracking the version wasn't fully in line with what the OTA updater really was responsible for.   I also assumed that the virtual_device would be more omnipresent than the OTA task, but it's possibly the case that we only really care about version when an OTA subsystem is present.  


>The rest of the VirtualTile infrastructure doesn't make any assumption that it will be reading a config from a json file or try to update it in any way so we're locking something in that preclude getting configuration from an in-memory dict or other mechanism.

I don't think this explicitly assumes there's a json configuration file -- the `self.config` always loads in these attributes from json now (with defaults if not present), but we could add a path or checker for using some in memory or `iotile registry`-like functionality alongside it. `describe_tile` as implemented has most attributes currently tied to default dummy values or config loaded files, so I would imagine something similar would have to change there too if that is the case.

mattrunchey: @timburke Does the above make sense? Like I said, the functionality can be moved to a particular task structure if it doesn't actually make sense to have it in the general `tile_based_device`. Let me know where you think it should be.

timburke: I guess my gut feeling is that it should be moved to a particular task instead of in the core virtual tile infrastructure.  I think we'll end up needing to make a lot of specific tweaks in it to pull/set the os an app tag information from the correct places, which may not always be the same place.  For example, we may want to have the os tag come from a file on the root hd so it's always up to date with whatever os is installed.

mattrunchey: I will put it in the task then, it definitely makes more sense if we're potentially going to support a more enumerated location/source set.

mattrunchey:
@timburke Took out all code referring to os/app info, and added an argument that disables the VirtualController for when we want to extend the behavior of the TileBasedDevice to have a more enhanced controller that doesn't require tweaking coretools. 

I also added the fix for event_loop in a separate commit within this PR since the change is relatively minor.

iotilecore/iotile/core/utilities/async_tools/event_loop.py
```diff
@@ -407,6 +407,19 @@ def inside_loop(self):
             if isinstance(result, Exception):
                 self._logger.error("Error stopping task %s: %s", task.name, repr(result))
 
+        # In some cases, tasks may not properly register or cancel their subtasks.
+        # This catches that in development, and also in deployment when there are third-party libraries
+        # that have this fault (such as hbmqtt)
+        tasks = [task for task in asyncio.Task.all_tasks() if task is not
+                 asyncio.tasks.Task.current_task() and not task.done()]
+        list(map(lambda task: task.cancel(), tasks))
```
timburke: I don't have a huge issue with this line but I've never seen it written that way before.  A list comprehension would also do it:

```python
[task.cancel() for task in tasks]
```

Issue #846, opened 2019-06-28
Add ability to flash firmware blob to bled112 dongles using coretools


timothyfroehlich:
At @ekkizogloy 's request, this should remove a significant amount of work that's currently being done manually.

The goal will be to run a python script that finds all bled112's and updates them. Will be using the transport-bled112 libraries as much as possible and the flashing code will likely need to live in `bled112_cmd.py`. I may see if I can integrate it into the main iotile code path (so you can do something like `iotile hw --port=bled112 install-blob --blob='file'`) But if it's non-trivial I'll just stick with the script.

Questions for @timburke:
Has this been attempted before? Is there anything I should know before I start? I've found the spec for how it should work.

What's going on with bled112_cmd.py and bled112_cmd.co.py? Which one should I use / modify? Both? I'm not familiar with asyncio yet (though I intend to learn soon) so I'd be more comfortable modifying the original.

timburke:
It is going to be very nontrivial to integrate this into `iotile` so I would just stick with a script.  You can bundle it with the `iotile-transport-bled112` package if you want as a `bled112-update-firmware` script.  

However, if you do that, you should not add any additional dependency requirements to the overall package requirements.  If additional dependencies are needed you should check for them in the script and tell the user to install them manually.  Most people using `iotile-transport-bled112` won't need to use the script so adding another installation dependency isn't warranted.

timburke:
`bled112_cmd.py` is fine to modify.  There was a partial port to support asyncio natively that is used when emulating a POD using `virtual-device`.  `bled112_cmd.py` is still used for client connections to PODs.

timothyfroehlich:
Sounds good. I'm not expecting to need any extra dependencies.

timburke:
Hmm, okay.  I'm surprised if you don't need a `libusb` dependency but maybe not.  I was always assuming the bled112 rebooted into USB DFU compatible mode for a firmware update and then you sent the blob using DFU.

timburke:
Nope, you're right.  It looks like it's all done over vcom:
![image](https://user-images.githubusercontent.com/2424759/60367754-e14ab780-99a3-11e9-9711-6144f2868b52.png)

timothyfroehlich:
Yep, exactly.

timothyfroehlich:
Well, maybe scratch all that? https://www.silabs.com/community/wireless/bluetooth/knowledge-base.entry.html/2015/08/06/_reference_updatin-Rtcr

```
[...]

PR #849, opened 2019-07-02
Add a new script to flash multiple Bluetooth bled112 dongles

timothyfroehlich:
## Overview

This adds a script to the transport/bled112 folder for flashing bled112 dongles. The script is independent of the rest of the iotile code, though uses the BLED112Adapter's find all dongles method.

### Additional Notes

- A re-packaged version of the firmware is included that is compatible with generic dfu flashing utilities.

Implements #846


timothyfroehlich:
Ready for review. @mattrunchey, have I committed any heinous Python crimes? @ekkizogloy does everything in the readme look good? It works like you'd expect?

timothyfroehlich:
@ekkizogloy Will merge in once you approve

Issue #852, opened 2019-07-11
Add "X/Y" completed count to bled112 flasher script


timothyfroehlich:
Report this count so that the operator can confirm that the expected number of dongles have appeared on the system.

Also add an "expected number" flag to the flasher, so that the operator can cause an error if the incorrect number of dongles appear.

@ekkizogloy

timothyfroehlich:
Also, add in a requirements.txt file for the script

PR #855, opened 2019-07-16
Make improvements and fixes to bluetooth flasher script

### LINKED ISSUE #853: Identify issue preventing bled112 flasher from working on production machine
There was a failure to import an exception observed, only in python 3.7. Using a python 3.6 environment worked fine.

I could not reproduce the issue on my machine with a python 3.7 install.

Need to confirm whether the python 3.7 environment on the production machine is standard.

@ekkizogloy

timothyfroehlich:
## Overview

Had some issues with the script, especially related to dongles failing to reset when requested to and missing python packages. This PR fixes all known issues / feature requests

## Major Changes

A more granular list of the changes made, ideally with links to issues

- Add an argument to specify the expected number of bled112 devices to find and then flash, error if count doesn't match. Implements #852 
- Add a requirements.txt file to ensure the correct dependencies are installed. Fixes #853 
- Fix issue where some dongles wouldn't reset when first requested (added retry to them)

@ekkizogloy

### DIFF
--- a/transport_plugins/bled112/blob/requirements.txt
+++ b/transport_plugins/bled112/blob/requirements.txt
@@ -1,0 +1,2 @@
+iotile-core>=5.0.0,<6
+pyserial>=3.4.0,<4
--- a/transport_plugins/bled112/blob/requirements.txt
+++ b/transport_plugins/bled112/blob/requirements.txt
@@ -1,3 +1,3 @@
 iotile-core>=5.0.0,<6
 pyserial>=3.4.0,<4
-iotile_transport_bled112
+iotile_transport_bled112>=3.0.0<4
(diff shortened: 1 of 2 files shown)


transport_plugins/bled112/blob/requirements.txt
```diff
@@ -0,0 +1,3 @@
+iotile-core>=5.0.0,<6
+pyserial>=3.4.0,<4
+iotile_transport_bled112
```
mattrunchey: Since you specified versions for the other two, we should have it present for the transport as well

mattrunchey: Though maybe this doesn't make sense, since the script is part of the transport plugin?

mattrunchey:
I wasn't sure if this is considered part of the bled112 transport plugin release or just a script inside of it, so didn't comment on release notes or versioning.

timothyfroehlich:
It's internal use only.

PR #854, opened 2019-07-15
Fix jlink multiple frame read

anwu:
### event_loop.py
- Allow `run_executor` to be run in coroutine finalizer function
### jlink.py
- Changed `disconnect` logic to close "rpc" interface last
- Added task to handle when iotile is `quit` to cleanup nicely
### jlink_background.py
- Changed frame reading logic from one at a time to all frames available in queue at once to minimize jlink memory read/writes
- Performance enhancement by removing delays
- Fixed coroutine quitting issues because of while loop in `maintenance_coroutine`

PR #856, opened 2019-07-16
Feat jlink upgrade

anwu:
## Main Changes
- Everything from https://github.com/iotile/coretools/pull/854
- Backwards compatibility check
- Added version to KNOWN_VERSIONS in control structure


iotilecore/RELEASE.md
```diff
@@ -2,6 +2,11 @@
 
 All major changes in each released version of `iotile-core` are listed here.
 
+## 5.0.3
+
+- JLink adapter support
```
timburke: I don't think there's any jlink specific change in here so that release note isn't needed unless I'm missing it.

transport_plugins/jlink/RELEASE.md
```diff
@@ -3,6 +3,10 @@
 All major changes in each released version of the jlink transport plugin are
 listed here.
 
+## 2.0.0
+
+- JLink adapter support for streaming and tracing
```
timburke: This should actually be `1.1.0` since from a public api point of view its 100% compatible.  The public API of a device adapter is defined as that it remains usable from `HardwareManager` and even though we rewrote the internals, from an outside pov it's just a feature update with support for streaming and tracing.

transport_plugins/jlink/iotile_transport_jlink/jlink.py
```diff
@@ -163,166 +189,195 @@ def _try_connect(self, connection_string):
...
-        """Enable script streaming interface for this IOTile device
+        self.opened_interfaces[interface] = True
 
-        Args:
-            conn_id (int): the unique identifier for the connection
-            callback (callback): Callback to be called when this command finishes
-                callback(conn_id, adapter_id, success, failure_reason)
-        """
+        if (interface == "tracing" or interface == "streaming") and self._control_info.version < 2:
+            logger.warning("Controller firmware does not support streaming/tracing.")
```
timburke: Just copying my other comment here from slack:

This should raise a `DeviceAdapterError` saying the interface is not supported on the jlink structure version being used.

anwu: Fixed.

transport_plugins/jlink/RELEASE.md
anwu: Fixed.

iotilecore/RELEASE.md
anwu: Keeping 5.0.3 version bump, removing JLINK note. Fixed.

timburke:
@anwu Based on the discussion in the `lib_controller` issue.  I think the remaining thing before we merge this and release an RC is to add in the retry logic on failing to acquire external gate.

If you want to merge this to master and then fix that in a separate PR to master, I'm fine or you can fix it and merge all together.  Either way.

We just shouldn't release the plugin as an rc until we have that issue resolved because the primary use case testing will be tracker_app usage which would run afoul of it.

anwu:
Linking issue https://github.com/iotile/lib_controller/issues/170 for historical purposes.

And yes I think I'll go ahead and resolve it with this PR before I merge.

anwu:
@timburke Added the retry logic, take a look at the latest commit and let me know what you think. I tested this out and saw it retry successfully. I saw the external gate attempt up to 3 times, so I chose the default retry as 5.

transport_plugins/jlink/iotile_transport_jlink/jlink.py
```diff
@@ -392,10 +392,20 @@ def add_trace(self, trace):
...
-            address, rpc_id, payload, timeout)
-        return response['payload']
+        response = None
+        
+        for attempt in range(retries):
+            try:
+                response = await self._jlink_async.send_rpc(
+                    self._device_info, self._control_info,
+                    address, rpc_id, payload, timeout)
+            except:
```
timburke: This should only catch `ExternalGate` errors since that's all we are retrying, it shouldn't chomp all RPC errors (such as timeouts or disconnections).  You may need to modify where the exception is thrown to distinguish it with a different internal class.

You should log with `debug` level as well when you retry since its expected to happen.

Finally if you get more than max retries attempts and don't get the gate, you should raise a `DeviceAdapterError` indicating failure, not just chomp the error.

anwu: Understood. I pushed a fix that reflects the behavior you just mentioned. So if it receives a `HardwareError` it will check if its an external gate error. If its not, it'll re-raise the exception higher. If it catches any other exception, it will just re-raise as well. Finally, when it finishes the loop after all the retries, it will raise a `DeviceAdapterError`.

transport_plugins/jlink/iotile_transport_jlink/jlink.py
```diff
@@ -331,48 +386,41 @@ def send_rpc_async(self, conn_id, address, rpc_id, payload, timeout, callback):
...
+
+        response = None
+
+        for attempt in range(retries):
+            try:
+                response = await self._jlink_async.send_rpc(
+                    self._device_info, self._control_info,
+                    address, rpc_id, payload, timeout)
+            except HardwareError as exc:
+                if exc.params['external_gate_error'] == 1:
```
timburke: This needs to be `exc.params.get('external_gate_error')` or it will raise KeyError when external_gate_error is not passed (so for any other HardwareError).  `get` from a dict will safely return `None` if the key is not found rather than raising.

anwu: Fixed.

PR #858, opened 2019-07-18
Add iotile-ship steps for firmware upgrades.

ekkizogloy:
## Overview

Add a few iotile-ship steps to flash firmware, set os/app tags and set the uuid of PODs.
These steps use the normal hardware manager / remote bridge interface via bluetooth to allow a 
ship script to be able to update firmware for a POD. 

## Major Changes

Addition of iotile/core/hw/debug/firmware_ship_steps.py which includes implementations of:
-  ReflashTileStep
-  SetFirmwareTagStep
-  SetUUIDStep

Modifies `setup.py` to add these as recipe steps.


anwu:
I just released a `5.0.2` . It was a really minor change, but I guess you'll have to rebase

ekkizogloy:
@anwu  This should have your 5.0.2 changes. in it.. I pulled and merged before submitting the PR. Let me know if something is missing.

iotilecore/setup.py
```diff
@@ -64,7 +64,10 @@
             'tile_based = iotile.core.hw.virtual.tile_based_device:TileBasedVirtualDevice'
         ],
         'iotile.recipe_action': [
-            'FlashBoardStep = iotile.core.hw.debug.flash_board_step:FlashBoardStep'
+            'FlashBoardStep = iotile.core.hw.debug.flash_board_step:FlashBoardStep',
+            'ReflashTileStep = iotile.core.hw.debug.firmware_ship_steps:ReflashTileStep',
+            'SetFirmwareTagStep = iotile.core.hw.debug.firmware_ship_steps:SetFirmwareTagStep',
+            'SetUUIDStep = iotile.core.hw.debug.firmware_ship_steps:SetUUIDStep'
```
timburke: We seem to already have the `SetUUIDStep` inside archtools that looks to be basically the same.  What's the advantage of duplicating it here vs fixing the other one if there's functionality missing?

https://code.eng.archsys.io/github.com/iotile/archtools/-/blob/archbuild/archbuild/recipe_steps/set_uuid_step.py

iotilecore/setup.py
```diff
@@ -64,7 +64,10 @@
             'tile_based = iotile.core.hw.virtual.tile_based_device:TileBasedVirtualDevice'
         ],
         'iotile.recipe_action': [
-            'FlashBoardStep = iotile.core.hw.debug.flash_board_step:FlashBoardStep'
+            'FlashBoardStep = iotile.core.hw.debug.flash_board_step:FlashBoardStep',
```
timburke: Is there a way these can all live in `archbuild` along with the other steps designed for internal manufacturing?  We ship `iotile-core` very broadly and I assume we already use archbuild other places in the manufacturing flow so it doesn't seem like it would be adding a new dependency.

Alternatively I would move it into `iotile-ship` so we can avoid including this code with every product that uses `iotile-core`:
https://code.eng.archsys.io/github.com/iotile/coretools/-/tree/iotileship/iotile/ship/actions

iotilecore/setup.py
ekkizogloy: Oh you are correct... I didn't realize it. I'll modify this one.

iotilecore/setup.py
ekkizogloy: I agree... I can move it into iotile-ship. I think that it would be better off there or archbuild...

mattrunchey:
@ekkizogloy -- did this end up getting put in to `archbuild`? Can we close this PR?

ekkizogloy:
Yes. Closing Now.

Issue #866, opened 2019-07-22
Add support for bundling data file with support wheels


timburke:
When `iotile-build` generates support wheels, it needs to check for the presence of non `.py` files indicating that data files are being bundled and generate the appropriate `setup.py` statements to include those files.

The files need to be accessible both when the component is installed in dev mode (`iotile registry add_component <PATH>`) as well as when the component pip installed.

All of the python wheel building happens in this file:
https://github.com/iotile/coretools/blob/master/iotilebuild/iotile/build/config/site_scons/pythondist.py

The end result of this issue should be that:
- if a module has a `support_package` product, then the folder containing the support package can contain non-python data files.  If this happens the built wheel will bundle those data files.

anwu:
Closed with https://github.com/iotile/coretools/pull/868

PR #868, opened 2019-07-23
Feat bundle data with support wheels

anwu:
## Overview

Allow data files to be bundled with its support wheel.
https://github.com/iotile/coretools/issues/866

## Major Changes

`iter_support_files` in `pythondist.py` inside of `iotilebuild`

- `if 'data' in os.path.normpath(dirpath).split(os.sep)` conditional to check if in a `data` directory/subdirectory, add file if it is (basically if `data` is explicitly in the directory path, add file)

### Testing

support ->data

image

### Additional Notes

- Still testing out a few more different nested directory structures


timburke:
You can add a basic unit test coverage following the pattern [here](https://github.com/iotile/coretools/tree/master/iotilebuild/test/test_iotilebuild)

You can do an `iotile build` on a basically empty component that just has a support wheel and a data file.  Then you can inspect the output wheel and ensure that it contains the data file using the technique here:
https://stackoverflow.com/a/32924572/9739119

anwu:
Added the basic test. The test checks for different nested directory structures. The files that end with `.ignore` should not end up in the output wheel. The only data files that make it in are the ones that end in `.use`.

iotilebuild/iotile/build/config/site_scons/pythondist.py
```diff
@@ -219,11 +227,20 @@ def generate_setup_py(target, source, env):
...
 
     data['entry_points'] = entry_points
+    data['include_package_data'] = True
 
     outdir = os.path.dirname(str(target[0]))
 
     render_template('setup.py.tpl', data, out_path=str(target[0]))
 
+    manifest_fn = outdir + '/MANIFEST.in'
+    with open(manifest_fn, 'w') as manifest_file:
```
anwu: Not sure if this is the preferred way to generate a file.

iotilebuild/iotile/build/config/site_scons/pythondist.py
```diff
@@ -219,11 +227,20 @@ def generate_setup_py(target, source, env):
         data['deps'] += tile.support_wheel_depends
 
     data['entry_points'] = entry_points
+    data['include_package_data'] = True
 
     outdir = os.path.dirname(str(target[0]))
 
     render_template('setup.py.tpl', data, out_path=str(target[0]))
 
+    manifest_fn = outdir + '/MANIFEST.in'
```
timburke:
```suggestion
    manifest_path = os.path.join(outdir, 'MANIFEST.in')
```

iotilebuild/iotile/build/config/site_scons/pythondist.py
```diff
@@ -219,11 +227,20 @@ def generate_setup_py(target, source, env):
...
+    with open(manifest_fn, 'w') as manifest_file:
+        for data in env['datafiles']:
+            manifest_file.write('include ')
```
timburke:
```suggestion
            manifest_file.write('include %s\n' % data)
```

iotilebuild/test/test_iotilebuild/python_data_bundling/SConstruct
```diff
@@ -0,0 +1,3 @@
+import autobuild
```
timburke: I see that you added a component with data files but I don't see any actual test case verifying that it builds correctly?

iotilebuild/iotile/build/config/site_scons/pythondist.py
```diff
@@ -61,7 +61,9 @@ def iter_support_files(tile):
     else:
         for dirpath, _dirnames, filenames in os.walk(support_package):
             for filename in filenames:
-                if not filename.endswith('.py'):
+                if 'data' in os.path.normpath(dirpath).split(os.sep):
```
timburke:
```suggestion
                if not filename.endswith('.py') and  'data' not in os.path.normpath(dirpath).split(os.sep):
                    continue
```

iotilebuild/iotile/build/config/site_scons/pythondist.py
```diff
@@ -154,10 +157,15 @@ def build_python_distribution(tile):
...
 
+        if 'data' in os.path.normpath(buildfile).split(os.sep):
+            datafile = os.path.join(tile.support_distribution, outpath)
+            datafiles.append(datafile)
+
         buildfiles.append(buildfile)
 
     # Create setup.py file and then use that to build a python wheel and an sdist
     env['TILE'] = tile
+    env['datafiles'] = datafiles
```
timburke: Since your `generate_setup.py` is building two files (`setup.py` and `MANIFEST.in`), you need to modify [line 174](https://github.com/iotile/coretools/pull/868/files#diff-7bf4f3fb3afa06b76d8a325626c25580R174) to include the MANIFEST.in file so that scons knows you're creating it.

You should also rename `generate_setup_py` to `generate_setup_and_manifest` so it's clear what it does.

iotilebuild/iotile/build/config/site_scons/pythondist.py
timburke: It's fine

iotilebuild/test/test_iotilebuild/python_data_bundling/SConstruct
anwu: Fixed. Added test case.

iotilebuild/iotile/build/config/site_scons/pythondist.py
anwu: Fixed.

anwu: Fixed.

iotilebuild/test/test_iotilebuild/python_data_bundling/SConstruct
anwu: Fixed. I don't know if you have a preference, but I added the test case to `test_scons.py`. I saw that a number of test components were doing that, and some had their tests in their individual `python/test`.

iotilebuild/test/test_iotilebuild/test_scons.py
```diff
@@ -381,3 +381,38 @@ def test_pytest(tmpdir):
...
+
+        assert os.path.exists(os.path.join('build', 'python', 'build', 'lib',
+            'iotile_support_py_data_bundle_1', 'data', 'file4.use'))
+
+        assert os.path.exists(os.path.join('build', 'python', 'build', 'lib',
+            'iotile_support_py_data_bundle_1', 'dir2', 'file6.ignore')) == False
+
+        assert os.path.exists(os.path.join('build', 'python', 'build', 'lib',
+            'iotile_support_py_data_bundle_1', 'dir2', 'dir2a', 'file7.ignore')) == False
+    finally:
```
timburke: Just one more thing to check.  You should verify that the built `wheel` also has the files just to close the loop.  You can check that using the technique here:
https://stackoverflow.com/questions/32923952/how-do-i-list-the-files-inside-a-python-wheel

anwu: Fixed. There was already a `list_wheel` function in `test_scons.py`, so I was able to just see if the expected files were in the list of wheel files with an `assert all(file in wheel_files for file in expected_files)`

timburke: Cool sounds good.

Issue #863, opened 2019-07-22
Create mock bled112 process for simulating high volume broadcast traffic.


timburke:
There is starter code for this [here](https://github.com/iotile/coretools/blob/master/transport_plugins/bled112/test/util/mock_bled112.py).  It is used internally for unit testing the bled112 device adapter.  

The goal of this issue should be a class that can generate high volume advertisement traffic and a command line script that takes a virtual serial port and acts like a bled112 dongle on that port.

Only a minimal amount of functionality is needed:
- set scan parameters: if active scan is true, then 2 advertisements are sent per "simulated device", one advertisement and one scan response packet, each 31 bytes.
- start scan
- stop scan
- query system.  This command is send when the bled112 device adapter turns on to figure out if any ble connections are active.

The command line script should log status periodically on what volume of broadcasts it is generating so that it's easy to see if the rate drops as we add additional things on the receiving side that could slow it down.

DimaWittmann:
Hi Tim! 
I am trying to make a complete picture in my head.

As I understood so far, I need to create a shell with a few commands (set parameters, start, stop, query).
As a argument this script accept path to device node.
I need to implement a class that will imitate a bled112 device that receives multiple adv packets (it jsut generate adv packets) and propagates it to gateway.

The gateway is expected to start and run independently of the script, we just know it should be connected to the other side of serial port.

The first question is how scan parameters should be propagated to gateway? It looks like gateway choose between active and passive scan.
**Start scan** cmd should make the mock to generate events with adv packets, provided that connection to bled adapter is established, shouldn't they?  
**Query system**, I haven't yet understood the purpose of this cmd. Bled device adapter initialize the connection to mock, start scan and than mock can only decide whether to send scan events or not. Please share any details if you can. 

Could you verify that my vision meets your expectation? I would be happy to see any additional thoughts  from you.

timburke:
I believe that's mostly in line with what I was thinking with the following caveats and some clarifications.

The `mock-bled112` script should be fully automatic in the sense that you start with a few command line arguments to specify how often to send advertisements (after the gateway calls `start scan`).  For example:

```
> mock-bled112 --max-advertisements-per-second 1000 --port /dev/side/b/of/virtual/tty
INFO: Starting emulated bled112 dongle serving 1000 messages per second
```

[...]

PR #869, opened 2019-07-30
Add scripts to test gateway throughput (#863)

DimaWittmann:
* script to profile gateway
* script to mock bled112 device
* add throttle-scan and throttle-scan-timeout config variables


transport_plugins/bled112/scripts/profile_gateway.py
```diff
@@ -0,0 +1,63 @@
...
+    loop.start()
+
+    yappi.start()
+
+    gateway = IOTileGateway(config, loop)
+    loop.run_coroutine(gateway.start())
+
+    hw = HardwareManager(port="ws:127.0.0.1:5120/iotile/v1")
+
+    time.sleep(time_to_profile)
```
timburke: This should be wrapped in a `try/except KeyboardInterrupt: pass` so that if you control C you can exit early cleanly.  As written it hangs the program because the main thread exits but it never cleans up the background event loop so python hangs at exit doing a `join` on the background thread.

transport_plugins/bled112/scripts/run_mock_observer.py
```diff
@@ -0,0 +1,48 @@
...
+        if data:
+            response = bled112.generate_response(data)
+            pts.write(response)
+
+    logging.info("Start sending advertisements")
+
+    while True:
+        current_time = time.time()
+
+        data = bled112.generate_multiple_adv_packets(args.max_advertisements_per_second)
```
timburke: When I use this according to the README, I get tons of the following log message:
image

It seems like there's a synchronization problem or something where maybe .  

The emulator may need to be a bit more complicated with foreground thread reading packets and responding to commands and a background thread sending advertisements while scanning is enabled.  You can use the `AsyncPacketBuffer` as an example of how it could be done.

transport_plugins/bled112/scripts/run_mock_observer.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+
+import argparse
+import time
+import signal
+import sys
+import logging
+import serial
+# TODO import properly, now one have to specify PYTHONPATH=../test
+from util.mock_adv_bled112 import MockAdvertisingBLED112
```
timburke: Ah, I see the issue now that you were talking about on the call.  I would move `MockAdvertisingBLED112` and its back class in `iotile_transport_bled112/hardware/emulator` and update the unit tests to find the mock there.  That removes the hack needed to import the test code here.

transport_plugins/bled112/scripts/profile_gateway.py
```diff
@@ -0,0 +1,63 @@
...
+    loop = BackgroundEventLoop()
+    loop.start()
+
+    yappi.start()
+
+    gateway = IOTileGateway(config, loop)
+    loop.run_coroutine(gateway.start())
+
+    hw = HardwareManager(port="ws:127.0.0.1:5120/iotile/v1")
+
```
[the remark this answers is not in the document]
DimaWittmann: Oh, I see, I check that scan returns expected result and was sure it reach the  HW.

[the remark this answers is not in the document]
DimaWittmann: Oh, I see, I checked that scan returns expected result and was sure packets reach the HW

transport_plugins/bled112/scripts/run_mock_observer.py
```diff
@@ -0,0 +1,48 @@
...
+    argparser.add_argument("--max-advertisements-per-second", type=int, required=True)
+    argparser.add_argument("--port", help="/dev/pts/X", required=True)
+    args = argparser.parse_args()
+
+    pts = serial.Serial(args.port, timeout=1)
+
+    bled112 = MockAdvertisingBLED112(3)
+
+    while not bled112.scanning:
+        data = pts.readline()
```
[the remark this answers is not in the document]
DimaWittmann: The approach was just to read as many as possible till the timeout fires, indeed on the best solution.
I will reuse async_packet for this purpose then.

transport_plugins/bled112/scripts/profile_gateway.py
DimaWittmann: I added try/except clauses and loop.stop() call

transport_plugins/bled112/scripts/run_mock_observer.py
DimaWittmann: Yes, on gateway restart there could be such a problem.
It looks like AsyncPacketBuffer resolves it.

transport_plugins/bled112/scripts/run_mock_observer.py
DimaWittmann: Thank you for the suggestion, I was not sure how to resolve it

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -544,6 +546,15 @@ def _process_scan_event(self, response):
         v2: There is only an advertisement and no scan response.
         """
 
+        info = None
+        # TODO make sure reading time in packets is of the same type
+        reading_time = time.time()
```
timburke: Since we only target python 3+ now, you should always use `time.monotonic()` when looking for a numeric value that can be used to determine timeouts.  Both `time.time()` and `time.monotonic()` return `floats` in unit of seconds but `time.monotonic()` is guaranteed to never decrease whereas time.time() report utc time so can have abrupt shifts if the kernel clock is updated or changes.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -569,21 +580,39 @@ def _process_scan_event(self, response):
...
+            info, reading_time, stream, reading = \
+                self._parse_v1_scan_response(string_address, data)
+
+        if info:
+            if self._throttle_broadcast and self._check_update_seen_broadcast(
+                        sender, reading_time, stream, reading,
+                        broadcast_toggle, counter=counter, channel=broadcast_multiplex):
+                return
+
+            self._trigger_callback('on_scan', self.id, info, self.ExpirationTime)
```
timburke: As discussed on the call, we should add a second config variables (`throttle_scans`) that controls whether these scans are throttled independently of the broadcast values.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
DimaWittmann: Okay, time.time() is used thought out the file

transport_plugins/bled112/iotile_transport_bled112/bled112.py
DimaWittmann: Yes, I added throttle-scan and throttle-scan-timeout to enable disable filtering of device_seen events and configure the timeout after which a same report or scan will be repeated

transport_plugins/bled112/iotile_transport_bled112/bled112.py
mattrunchey: Yea, we should probably update all of those with `time.monotonic()`

DimaWittmann: I see, I updated those in bled112.

DimaWittmann:
Hi @timburke! 
Just to let you know, I updated the PR

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -69,6 +69,8 @@ def __init__(self, port, on_scan=None, on_disconnect=None, passive=None, **kwarg
             self._active_scan = config.get('bled112:active-scan')
 
         self._throttle_broadcast = config.get('bled112:throttle-broadcast')
+        self._throttle_scans = config.get('bled112:throttle-scan')
+        self._throttle_scans_timeout = config.get('bled112:throttle-scan-timeout')
```
timburke: Since this applies to both scan and broadcasts it should just be called `bled112:throttle-timeout`

transport_plugins/bled112/iotile_transport_bled112/config_variables.py
```diff
@@ -7,5 +7,7 @@ def get_variables():
     conf_vars = []
     conf_vars.append(["active-scan", "bool", "Probe devices during scan for uptime, voltage and broadcast data", "false"])
     conf_vars.append(["throttle-broadcast", "bool", "Only report changing broadcast values, not all values", "false"])
+    conf_vars.append(["throttle-scan", "bool", "Only report device_seen once in a timeout", "false"])
+    conf_vars.append(["throttle-scan-timeout", "int", "device_seen and report events timeout in seconds", 30])
```
timburke: Same rename

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -569,21 +579,40 @@ def _process_scan_event(self, response):
...
             if data[22] == 0xFF and data[23] == 0xC0 and data[24] == 0x3:
                 self._v1_scan_count += 1
-                self._parse_v1_advertisement(rssi, string_address, data)
+                info = self._parse_v1_advertisement(rssi, string_address, data)
             elif data[3] == 27 and data[4] == 0x16 and data[5] == 0xdd and data[6] == 0xfd:
                 self._v2_scan_count += 1
-                self._parse_v2_advertisement(rssi, string_address, data)
+                info, reading_time, stream, reading, \
+                    broadcast_toggle, counter, broadcast_multiplex = \
+                    self._parse_v2_advertisement(rssi, string_address, data)
```
mattrunchey: This is going to fail when you return as you do on line 615. There, you specified 4 Nones, but you are unpacking 7 values. Example:
```
>>> def f(a=0):
...   if a == 0:
...     return None, None, None
...   return 1, 2, 3, 4, 5
... 
>>> a,b,c,d,e = f()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected 5, got 3)
```

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -569,21 +579,40 @@ def _process_scan_event(self, response):
...
+        if info:
+            drop_broadcast = self._check_update_seen_broadcast(
+                sender, reading_time, stream, reading,
+                broadcast_toggle, counter=counter, channel=broadcast_multiplex)
+
+            if not self._throttle_scans or not drop_broadcast:
+                self._trigger_callback('on_scan', self.id, info, self.ExpirationTime)
+
+            # If there is a valid reading on the advertising data, broadcast it
+            if (not self._throttle_broadcast or not drop_broadcast) and \
```
mattrunchey: The way this logical condition is written seems more confusing than necessary.
```suggestion
            if not (self._throttle_broadcast and drop_broadcast) and \
```

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -569,21 +579,40 @@ def _process_scan_event(self, response):
...
+                report = BroadcastReport.FromReadings(info['uuid'], [io_tile_reading], reading_time)
+                self._trigger_callback('on_report', None, report)
+
 
     def _parse_v2_advertisement(self, rssi, sender, data):
         """ Parse the IOTile Specific advertisement packet"""
 
         if len(data) != 31:
-            return
+            return None, None, None, None
```
mattrunchey: Can we get some comments or record of the changes we made to this function (`_parse_v2_advertisement`) ? It's a pretty core functionality and I don't follow why the return type here is 4 nones, and the return further on in the function gives back 7 values (and doesn't do any callbacks). 

Why did the callbacks get moved to the `_process_scan_event` function?

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -707,16 +728,7 @@ def _parse_v1_scan_response(self, sender, scan_data):
...
-
-        # If there is a valid reading on the advertising data, broadcast it
-        if stream != 0xFFFF:
-            if self._throttle_broadcast and self._check_update_seen_broadcast(sender, reading_time, stream, reading):
-                return
-
-            reading = IOTileReading(reading_time, stream, reading, reading_time=datetime.datetime.utcnow())
-            report = BroadcastReport.FromReadings(info['uuid'], [reading], curr_time)
-            self._trigger_callback('on_report', None, report)
+        return info, reading_time, stream, reading
```
mattrunchey: Same comment here. We should document why it was necessary to factor out this functionality and be very careful about changing logic around this part of the code.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
DimaWittmann: Thank you for pointing out, this I missed!

transport_plugins/bled112/iotile_transport_bled112/bled112.py
DimaWittmann: Agreed!

transport_plugins/bled112/iotile_transport_bled112/bled112.py
DimaWittmann: I moved handling of callback up to _process_scan_event to remove code duplication.  
In case you see more harm that good or the logic is confusing, I can revert this changes and copy  call back handling in every parsing function, but still it will be changes in core functionality.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
DimaWittmann: It's not necessary, my intention was  to handle callback in one place.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
DimaWittmann: Yes, it looks better

transport_plugins/bled112/iotile_transport_bled112/bled112.py
mattrunchey: Got it, thanks. I understand that we're still changing functionality with the debug path, but it's crucial that we keep a record of what changed here (preferably in the commit or RELEASE comments) in case we do need to roll something back that we didn't catch. The refactor to group repeated code is fine, it just seemed to be a change that wasn't a required functionality at first glance so I needed to mention it.

DimaWittmann: I added a notice to commit message, thank you for pointing this out.

mattrunchey:
Hey @DimaWittmann -- sorry for leaving this here for so long. Would you consider it ready to merge? I can resolve the conflicts with the main branch and close this out, if so - let me know.

DimaWittmann:
Hi @mattrunchey, thank you for the message. You can proceed with merging

PR #871, opened 2019-07-31
Feat elf analyzer

anwu:
## Overview

Created a tool to let people test/verify `.elf` images.

## Major Changes

No changes to any existing code. This is a new utility class that shouldn't affect any preexisting code.

The tool simply uses `IntelHex` to get information about an `.elf` file, and scans the firmware image for a magic number to locate its `cdb_app_info`.

### Additional Notes

- Future plans to incorporate verifying devices with loaded images with image files

### DIFF
--- a/iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
+++ b/iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
@@ -3,9 +3,13 @@ from iotile.core.exceptions import *
 from tempfile import NamedTemporaryFile
 import os
 import subprocess
+import struct
 
 class FirmwareImageAnalyzer:
     """A firmware image analyzer using IntelHex"""
+
+    KNOWN_FIRMWARE_NAMES = ['boot52', 'NRF52 ', 'firmPP', 'UART  ', 'firmFF']
+
     def __init__(self, firmware):
         if not firmware.endswith(".elf") and not firmware.endswith(".hex"):
             raise ArgumentError("You must pass an ARM firmware image in elf/hex format", path=firmware)
@@ -20,7 +24,7 @@ class FirmwareImageAnalyzer:
                 err = subprocess.call(['arm-none-eabi-objcopy', '-O', 'ihex', firmware, tmp])
                 if err != 0:
                     raise ExternalError("Cannot convert elf to binary file", error_code=err)
-            
+
                 hex_image = IntelHex(tmp)
             else:
                 hex_image = IntelHex(firmware)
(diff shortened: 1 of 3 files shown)


iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
```diff
@@ -0,0 +1,155 @@
...
+        self.min_addr = self._hex_image.minaddr()
+        self.max_addr = self._hex_image.maxaddr()
+        self.num_addresses = len(self._hex_image.addresses())
+        self.segments = self._hex_image.segments()
+        self.memory_size = self._hex_image.get_memory_size()
+        self.magic_number = magic_number
+
+    def set_firmware_magic_number(self, magic_number):
+        """Sets the magic number to search a CDB app info block by
+        """
```
mattrunchey: Single line comments should have closing quotes on the same line

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def _format_cdb_app_info(self, cdb_app_info_bytes):
+        """Helper function to format the cdb_app_info byte string to dict
+        """
+        hardware_type, api_major_version, api_minor_version, name, module_major_version, \
+            module_minor_version, module_patch_version, num_slave_commands, num_required_configs, \
+            num_total_configs, reserved, p_config_variables, p_slave_handlers, magic_number, \
+            firmware_checksum = struct.unpack("<BBB6sBBBBBBBLLLL", cdb_app_info_bytes)
+
+        cdb_app_info = dict()
```
mattrunchey: This is a valid way to populate the dictionary, but you can also initialize it in place with something like

```python
cdb_app_info = { 'hardware_type': hardware_type,
                             'api_major_version': api_major_version,
```
                            ...
                           }

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
anwu: Fixed

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
anwu: Fixed

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
```diff
@@ -0,0 +1,148 @@
...
+        """Helper function to check the byte string if it is a valid CDB app info"""
+        _hardware_type, _api_major_version, _api_minor_version, name, _module_major_version, \
+            _module_minor_version, _module_patch_version, _num_slave_commands, _num_required_configs, \
+            _num_total_configs, _reserved, _p_config_variables, _p_slave_handlers, magic_number, \
+            _firmware_checksum = struct.unpack("<BBB6sBBBBBBBLLLL", cdb_app_info_bytes)
+        cdb_app_info = self._format_cdb_app_info(cdb_app_info_bytes)
+
+        try:
+            name = cdb_app_info['name'].decode("utf-8")
+        except:
```
mattrunchey: Can you comment on why this exception should be passed? Looking at the try, it isn't entirely clear why failing to decode should just return False. I'm assuming it's just to inform the user that their input wasn't valid, but in that case, should we not raise an exception to that extent and handle it in the caller?

If you choose to eat this exception, pylint will complain about it (because we usually want to raise the exception to see what went wrong further up the call chain), so you typically want to annotate it:
```suggestion
        except:  # pylint:disable=bare-except;reason we are using this anyways
```

mattrunchey: It also overlaps with the last return False in this function, which is returning false for a different reason -- so I would imagine that there are two different modes of `return False` that are being bundled here (the second one being that it actually doesn't pass the conditional, and the first one being that it could have failed to decode the bytes)

anwu: @mattrunchey So some background knowledge... the CDB app info block looks like the following structure:
```
typedef struct
{
    uint8_t                 hardware_type;
    uint8_t                 api_major_version;
    uint8_t                 api_minor_version;

    char                    name[kCDBModuleNameLength];

    uint8_t                 module_major_version;
    uint8_t                 module_minor_version;        
    uint8_t                 module_patch_version;

    uint8_t                 num_slave_commands;
    uint8_t                 num_required_configs;
    uint8_t                 num_total_configs;
    
    uint8_t                 reserved;

    const cdb_config_entry  *config_variables;
    const cdb_slave_entry   *slave_handlers;        

    uint32_t                magic_number;
    uint32_t                firmware_checksum;
} CDBApplicationInfoBlock;
```
And when we build our controller, we include one of these at the end of our firmware. Some legacy controller images may have it in different locations of our firmware.

So in `get_cdb_app_info`, I first check the most common location where the app info might be -- at the end of the image. So to verify ( `_check_cdb_app_info`) that the location I'm looking at is indeed a `cdb app info`, I take a look at its magic number, along with the name. The name needs to be a string (or char array) that is of one of the recognizable firmware names. If the `name` in that location is just some random string of bytes that are not `utf-8`, it will throw an error. 

So tldr;
`_check_cdb_app_info` is supposed to verify a block of memory is a `cdb app info` block by checking the name and magic number. Ideally, it would look like 
` if cdb_app_info['name'].decode("utf-8") in KNOWN_FIRMWARE_NAMES and cdb_app_info['magic_number'] == self.magic_number`, but `cdb_app_info['name'].decode("utf-8")` can error if its not a decodeable string, thus making that memory block NOT a `cdb app info`.

anwu: Added better exception handling.

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
```diff
@@ -0,0 +1,150 @@
...
+            raise ExternalError("Unexpected error resolving the firmware name.")
+
+        if name in FirmwareImageAnalyzer.KNOWN_FIRMWARE_NAMES \
+            and cdb_app_info['magic_number'] == self.magic_number:
+            return True
+
+        return False
+
+    def _format_cdb_app_info(self, cdb_app_info_bytes):
```
timburke: Style points.  This can just be a standalone function with the same name at the bottom of the module.  Standalone functions can be better because they're easier to test (you don't need to create the class, just call the function) and from an understanding point of view, they make the class smaller so it's easier to get a sense of what it does.

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
```diff
@@ -0,0 +1,150 @@
...
+from iotile.core.exceptions import *
+from tempfile import NamedTemporaryFile
+import os
+import subprocess
+import struct
+
+class FirmwareImageAnalyzer:
+    """A firmware image analyzer using IntelHex"""
+
+    KNOWN_FIRMWARE_NAMES = ['boot52', 'NRF52 ', 'firmPP', 'UART  ', 'firmFF']
```
timburke: I wouldn't hardcode these names here.  It just means that we will have to update this package whenever we add another name and as I see the intended use-case, it's just a secondary check that we've found the info block.  I don't think that secondary check is worth the hassle of needing to keep this list in check with some internal list of firmware names.

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
```diff
@@ -0,0 +1,150 @@
...
+    SIZE_OF_CDBAPPINFO = 32
+    MAGIC_NUMBER_BYTE_OFFSET = 24
+
+    def __init__(self, firmware, magic_number=0xBAADDAAD):
```
timburke: I would recommend making this class a bit more immutable so that it's easier to work with:

- You already have a `magic_number` option in `__init__`, so there's not a need to allow the magic number to be set later.   The user can just create the class when once they know the magic number they want to use.
- This means all of the analysis can be done immediately in `__init__` since the possibility of the magic number changing was the only thing I can see that lead to deferring the analysis until later.
- This means that the class can just have a declarative `app_info` property that contains the results.  If analysis wasn't possible, it can raise a `DataError` from the `__init__` method.

This should simplify the class and remove some edge cases that currently are broken.  For example, you are allowed currently to call `set_firmware_magic_number` after initial construction but if you call it after the first time you have called any accessor function, it will be silently ignored because the results will have already been cached.

Moving everything to happen during `__init__` makes it more clear.  You create the class to analyze a firmware and then you can access the results.  If you want to analyze the firmware differently, create another instance of the class with different arguments.

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
anwu: Fixed in next push.

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
anwu: This is true. However, I ran into the case where I saw the magic number elsewhere in the firmware where it was *not* the cdb app info block. That is why I added this secondary check. I'm not too sure what other parameters I can check within the `CDBApplicationInfoBlock`. Maybe I can just attempt to `.decode("utf-8")` to make sure that the `name` is in fact a string?

timburke: That's one potential option.  An alternative is to validate the checksum embedded in the application block against the firmware image itself.  This second option though won't work for legacy controller firmwares but will work for all tile firmware.

I would probably recommend just verifying that the name is a utf-8 encoded string and that should be good enough.

anwu: Got it, fixed in next push.

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
anwu: Fixed in next push.

timburke:
> Looks good to me, however I will let @timburke respond regarding including this tool in coretools proper or should it be moved to the archtools repo.

Generally I like these things living in archtools but this is needed by `iotile-support-lib-controller` and we don't want to add a dependency on `archtools` to that package.

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
```diff
@@ -0,0 +1,140 @@
...
+    def get_firmware_hardware_type(self):
+        """Returns hardware type specified in cdb app info"""
+        if self.cdb_app_info is None:
+            self.get_cdb_app_info()
+
+        return self.cdb_app_info['hardware_type']
+
+    def get_firmware_api_version(self):
+        """Returns firmware api version specified in cdb app info"""
+        if self.cdb_app_info is None:
```
timburke: You can get rid of all of these checks since you know if the object was created you have a `cdb_app_info`.  That's one of the key benefits of doing processing during `__init__`.  A user doesn't get an instance unless the __init__ processing was successful.

anwu: You right. Fixed.

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def _check_cdb_app_info(self, cdb_app_info_bytes):
+        """Helper function to check the byte string if it is a valid CDB app info"""
+        cdb_app_info = format_cdb_app_info(cdb_app_info_bytes)
+
+        try:
+            cdb_app_info['name'].decode("utf-8")
+        except UnicodeDecodeError:
+            return False
+        except:
```
mattrunchey: You still need to justify this exception. Why are you raising an ExternalError instead of just letting the original exception raise? Is the caller expected to handle an `ExternalError` instead of whatever does happen? 

If that's the case, you need to put a `pylint=disable-bare-except` clause on this to supress the pylint warning. 

Unless we never want to fix the cause of the ExternalError, we shouldn't be overwriting what the actual error is. If you *do* need to raise it instead of the original one, you should at least log the exception and stacktrace.

timburke: One option is:

```
except Exception as err:
    raise ExternalError("Unknown whatever") from err
```

That fully preserves the original exception information and stacktrace so it can be debugged.

anwu:
Feature moved to internal firmware repo...

iotilecore/iotile/core/hw/debug/firmware_image_analyzer.py
anwu: Added Tim's implementation.

PR #872, opened 2019-07-31
Add support for passing kwargs down to Api() object.

timburke:
This PR adds support to `iotile-ext-cloud` to pass additional arguments to the `iotile-cloud` Api() object that is used to make rest calls to iotile.cloud.  This is needed to be able to configure the `timeout` and `retries` parameters that were added to control what happens in the case of transient network failures.

Previously, `iotile-cloud` didn't support automatic retries on connection errors, now it does so we need this change to be able to use that ability in CoreTools.

See https://github.com/iotile/python_iotile_cloud/pull/62

This is an urgent hotfix to resolve a stability issue with an uploader running inside a network that periodically chooses to drop connections to iotile.cloud.

### DIFF
--- a/iotile_ext_cloud/iotile/cloud/cloud.py
+++ b/iotile_ext_cloud/iotile/cloud/cloud.py
@@ -3,10 +3,9 @@
 from io import BytesIO
 import getpass
 import datetime
+from collections import namedtuple
 from dateutil.tz import tzutc
 import dateutil.parser
-from collections import namedtuple
-import requests
 
 from iotile_cloud.api.connection import Api
 from iotile_cloud.api.exceptions import RestHttpBaseException, HttpNotFoundError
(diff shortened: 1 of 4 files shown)


iotile_ext_cloud/iotile/cloud/cloud.py
```diff
@@ -159,7 +158,7 @@ def get_whitelist(self, device_id):
     def check_time(self, max_slop=300):
         """ Check if current system time is consistent with iotile.cloud time"""
 
-        cloud_time = requests.get('https://iotile.cloud/api/v1/server/').json().get('now', None)
+        cloud_time = self.api.session.get('https://iotile.cloud/api/v1/server/').json().get('now', None)
```
dkarchmer: I have forgotten about this API, which I better port to ArchFX Cloud.
Side note: We will need a way for cloud_ext not to hard code iotile.cloud as it should also be able to talk to archfx.io.
Shouldn't this be controller by the iotile config set cloud:server setting?

timburke: Good catch.  This is the only place it's hard-coded I think but I will verify.  There is already a configurable domain as you say, it looks like this just was a bit hacky.  It's never used in production code anymore I think but I'll fix it before merging.

PR #873, opened 2019-08-07
Feat support ble multiplexed broadcast

anwu:
## Overview

Add support on `coretools` stack to allow BLE multiplexed broadcast channels.

## Major Changes

Instead of treating each UUID as its own broadcast item, we will now treat each STREAM belonging to each UUID as its own broadcast item. This lets us display multiple broadcast streams for each UUID.

image

### Additional Notes

At first I modified `linebuffer_ui.py`, but other functions like `watch_scans` were affected by the change. By containing the change in `hwmanager`'s `watch_broadcasts`, we won't break other tools using `linebuffer_ui`.

### DIFF
--- a/iotilecore/RELEASE.md
+++ b/iotilecore/RELEASE.md
@@ -2,6 +2,10 @@
 
 All major changes in each released version of `iotile-core` are listed here.
 
+## 5.0.6
+
+- Modify `watch_broadcasts` to treat individual UUID's streams as different items
+
 ## 5.0.5
 
 - Fix TileBasedDevice controller override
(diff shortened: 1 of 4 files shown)


mattrunchey:
Exciting to see that we may be able to send multiple streams of broadcast data! Agree that we can just keep this in HEAD unless there is an urgent need for someone that's not a dev working from master.

iotilecore/RELEASE.md
```diff
@@ -2,6 +2,10 @@
 
 All major changes in each released version of `iotile-core` are listed here.
 
+## 5.0.6
```
mattrunchey: Let's call this HEAD and not create a new version

anwu: Got it

PR #874, opened 2019-08-21
fix: add trailing Z when isoformatting datetime

lelithium:
## Overview

Adds a trailing Z to serialized timestamps in `FlexibleDictionaryReport` 

### Additional Notes

- Adds a `tzinfo` marker when using `dateutil.parser.parse`


lelithium:
I meant a change in the parsed object (namely, a `tzinfo` attr is now parsed from the trailing `Z`, whereas it wasn't before)

PR #875, opened 2019-09-05
Fix improve forensic dump

anwu:
## Overview

Improvement on flash forensics tool.
Previously, the user had to do
`dump_memory --out_path dump.bin --start_addr 0 --data_length 0x1000 --memory_region ram`
With this change, the user just needs to do
`dump_memory dump.bin --memory_region ram`
to dump all of ram.
The corresponding call in `iotile_transport_jlink` already checks that if there is no specified length, that it will dump all of that memory region.

## Major Changes

Give `dump_memory` default parameters.


anwu:
I made some additional changes. The changes I made previously improved usage for memory types other than external memory.

The more recent changes improve usage for external memory dumping. The improvement made depends on https://github.com/iotile/con_btc1/pull/196 's executable it produces.
### Main Changes
- `flash_forensics_config.py` added addresses to read new response, also updated program start address
- Fixed `_inject_blob`, previous JLink upgrade did something wonky to it
- `_write_ff_query_id_cmd` just writes the command id to where the injected blob is looking for commands
- Patched up a hole in `_write_ff_dump_cmd`. Previously it just assumed to dump memory, since it was the only command, now we have a query command
- `_read_ff_query_size_resp` checks the JEDEC ID against the two known SPI chips we use
- `_read_mapped_memory` allows user to not input specific start address/data length, instead, it will query the device's information to find its max size to just dump the entire chip

PR #878, opened 2019-09-25
Fix broadcast report bug that caused double callbacks

### LINKED ISSUE #877: Something is causing us to see double BroadcastReports
We have a few PODs set up to broadcast at an interval of 100ms. This interval was confirmed by a 3rd party app (NRF Connect). As a result, we expect to see a maximum of 10 unique broadcast packets per second when we are looking at them here (with `watch_broadcasts` or `iter_broadcast_reports`). 

However, we are seeing around 15 broadcasts per second. Here is a snippet of what the contents of the IOTileReadings look like:
```
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.021661', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.021661', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.133200', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.133200', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.242691', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.242691', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.265342', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.265342', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.377426', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.377426', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.486444', 'value': 8, 'streamer_local_id': 0}
{'device_timestamp': 4970, 'stream': 4136, 'timestamp': '1970-01-04T19:26:40.486444', 'value': 8, 'streamer_local_id': 0}
```

The key thing to point out here is that each IOTileReading seems to be duplicated and has the exact same `timestamp` field. In reality, we're seeing around 7.5 broadcasts per second -- but are parsing 15 due to duplication.

It's not clear where this duplication is happening. The behavior is consistent regardless of the `active-scan` setting, and `BroadcastReport` is only used in two places in the BLED112 transport plugin code (in https://github.com/iotile/coretools/blob/master/transport_plugins/bled112/iotile_transport_bled112/bled112.py#L637 and https://github.com/iotile/coretools/blob/master/transport_plugins/bled112/iotile_transport_bled112/bled112.py#L718 ). 

If you want to do any verification of this, you can use something similar to the following script to print out the readings of any specific device:

```python
from iotile.core.hw import HardwareManager
from datetime import datetime
import asyncio 
import time

device_map = {}

def _handle_broadcast(report):
    device = report.origin
    if device not in device_map:
        device_map[device] = {'num_broadcasts': 1}
    else:
        device_map[device]['num_broadcasts'] += 1
    
    if device == 2964:
        print(report.decode()[0][0].asdict())

with HardwareManager(port='bled112') as hw:
    hw.enable_broadcasting()
    starttime = datetime.now()
    while 1:
        try:
            for report in hw.iter_broadcast_reports(blocking=False):
                _handle_broadcast(report)
            time.sleep(0.5)
        except (asyncio.CancelledError, KeyboardInterrupt):
            print("Task Cancelled by Ctrl+C")
            break
        
        checktime = datetime.now()
        print('0xB94 : ' + str(device_map[2964]['num_broadcasts']/(checktime-starttime).seconds))
        #for device in device_map:
        #    print(str(device) + ": " + str(device_map[device]['num_broadcasts']/(checktime-starttime).seconds)) 

```

mattrunchey:
Fix logging error in async task shutdown logic

## Overview

Fixes #877 

New release needed so that we can add this to our node release for better performance.

### DIFF
--- a/iotilecore/RELEASE.md
+++ b/iotilecore/RELEASE.md
@@ -2,10 +2,12 @@
 
 All major changes in each released version of `iotile-core` are listed here.
 
-## HEAD
+## 5.0.6
 
 - Modify `watch_broadcasts` to treat individual UUID's streams as different items
 - Modify parameters for `dump_memory` to use default addresses/lengths
+- Fix double passing of BroadcastReports
+- Fix logging error on task cancellation
 
 ## 5.0.5
 
--- a/iotilecore/iotile/core/hw/transport/adapter/mixin_notifications.py
+++ b/iotilecore/iotile/core/hw/transport/adapter/mixin_notifications.py
@@ -175,11 +175,11 @@ class BasicNotificationMixin:
             self._currently_notifying = True
             conn_id = self._get_conn_id(conn_string)
 
-            event_maps = self._monitors.get(conn_string, {})
-            wildcard_maps = self._monitors.get(None, {})
-
-            wildcard_handlers = wildcard_maps.get(name, {})
+            event_maps = self._monitors.get(conn_string, {}) if conn_string is not None else {}
             event_handlers = event_maps.get(name, {})
+
+            wildcard_maps = self._monitors.get(None, {})
+            wildcard_handlers = wildcard_maps.get(name, {})
 
             for handler, func in itertools.chain(event_handlers.items(), wildcard_handlers.items()):
                 try:
--- a/iotilecore/iotile/core/utilities/async_tools/event_loop.py
+++ b/iotilecore/iotile/core/utilities/async_tools/event_loop.py
@@ -405,7 +405,7 @@ class BackgroundEventLoop:
         results = await asyncio.gather(*awaitables, return_exceptions=True)
         for task, result in zip(self.tasks, results):
             if isinstance(result, Exception):
-                self._logger.error("Error stopping task %s: %s", task.name, repr(result))
+                self._logger.error("Error stopping task %s: %s", task, repr(result))
 
         # In some cases, tasks may not properly register or cancel their subtasks.
         # This catches that in development, and also in deployment when there are third-party libraries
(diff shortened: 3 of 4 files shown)


x24ling:
If you're curious, these are the reports/second values I get from the PODS in the farm with the changes in the PR when the sensor_graph sets the broadcast period to 100ms:

```
0x81d: 7.462809917355372
0x816: 7.396694214876033
0x81b: 6.024793388429752
0x815: 6.975206611570248
0x817: 7.347107438016529
0x81e: 7.570247933884297
0x81c: 7.190082644628099
0x81a: 7.363636363636363
0x81f: 7.305785123966942
```

Issue #876, opened 2019-09-13
Add support for matching proxies to tiles using firmware version


timburke:
Currently, we only match proxies to tiles if the module name matches.  If multiple proxies are installed with the same module name, the "first one" loaded will win always.

However if you install 2 versions of a tile support package, say:

```
iotile-support-con-nrf52832-3
iotile-support-con-nrf52832-4
```

For two different major firmware versions, it should be able to use the `ModuleVersion()` information from the proxy to restrict what tile it matches.  

Most existing proxies won't have a ModuleVersion() classmethod, which should be interpreted to mean that they match all firmware versions for backwards compatibility with the current behavior.

**Notes**

- The key place to modify is [here](https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/hw/hwmanager.py#L779)
- You should expect `ModuleVersion`, if present in the proxy, to return a list of `SemanticVersionRange` objects which are used to match the firmware version loaded from the tile

anwu:
@timburke Your link pointed me to a slack archive. Did you mean to point somewhere in `proxy.py` in coretools?

anwu:
Closes https://github.com/iotile/lib_controller/issues/199

timburke:
@anwu Thanks, https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/hw/hwmanager.py#L779

anwu:
> @anwu Thanks, https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/hw/hwmanager.py#L779

I've been looking through the code and I think [this ](https://code.eng.archsys.io/github.com/iotile/coretools@8cf93381d69dcfc6768885420cfd0d8b7881e771/-/blob/iotilecore/iotile/core/hw/proxy/external_proxy.py#L12)is the correct place to modify.
The location you linked deals with actual tile proxies. Where the location I linked deals with proxy plugins that you get from support packages I think.

timburke:
There are a couple of things going on here that need to be disentangled:

1. How proxy classes are matched to tiles.
2. How the controller proxy finds "proxy plugins" from `lib_controller`.

### Matching Proxies to Tiles

This process is governed by the location in `HardwareManager` that I linked to and needs to be updated to support including `ModuleVersion()` in the match criteria, when specified.

### Assembling the controller proxy

[...]

PR #880, opened 2019-10-01
Proxy matching and ReferenceControllerProxy

anwu:
## Overview

Implementing proxy matching when obtaining tile proxy.
Also implemented the use of the `ReferenceControllerProxy` from https://github.com/iotile/lib_controller/pull/203.

## Major Changes

- Test suites in `iotileemulate` have been updated to use `ReferenceControllerProxy`
- Had to set `basic=False` in `hw.get` calls to be able to inherit properly

### Additional Notes

- Not complete yet! Just beginning a draft. Still need to implement proxy matching.
Closes https://github.com/iotile/coretools/issues/876


iotilecore/iotile/core/hw/hwmanager.py
```diff
@@ -785,7 +788,47 @@ def get_proxy(self, short_name):
...
+            except AttributeError:
+                # If proxy does not have ModuleVersion(), use the name of the support package
+                stripped_proxy_version = str(proxy).split('.')[0].split('_')[-1]
+                module_version = SemanticVersionRange.FromString('^' + stripped_proxy_version + '.0.0')
+                proxy_details[proxy]['ModuleVersion'] = module_version
+
+            # Keep track of each proxy's minimum version required
+            proxy_details[proxy]['MinVersion'] = module_version._disjuncts[0][0][0]
+
+            if module_version.check(tile_version):
```
mattrunchey: Make sure `proxy.ModuleVersion()` returns a `SemanticVersionRange`, otherwise I'm not sure this check will work.

iotilecore/iotile/core/hw/hwmanager.py
```diff
@@ -785,7 +788,47 @@ def get_proxy(self, short_name):
...
+                proxy_details[proxy]['Compatible'] = True
+                min_version = proxy_details[proxy]['MinVersion']
+                latest_version = min_version if latest_version.__lt__(min_version) else latest_version
+            else:
+                proxy_details[proxy]['Compatible'] = False
+
+        # Return the latest/compatible proxy
+        for proxy, details in proxy_details.items():
+            if details['Compatible'] and details['MinVersion'] == latest_version:
+                return proxy
```
mattrunchey: I think you will need to iterate through the whole list and use a clamp (like recording the biggest number you've seen so far in a list) to select the "latest", as it is this returns on the first match instead of the latest match.

iotilecore/iotile/core/hw/hwmanager.py
```diff
@@ -785,7 +788,47 @@ def get_proxy(self, short_name):
...
+            proxy_details[proxy] = {}
+            try:
+                # If proxy has ModuleVersion(), get the SemanticVersionRange
+                module_version = proxy.ModuleVersion()
+                proxy_details[proxy]['ModuleVersion'] = module_version
+            except AttributeError:
+                # If proxy does not have ModuleVersion(), use the name of the support package
+                stripped_proxy_version = str(proxy).split('.')[0].split('_')[-1]
+                module_version = SemanticVersionRange.FromString('^' + stripped_proxy_version + '.0.0')
+                proxy_details[proxy]['ModuleVersion'] = module_version
```
mattrunchey: I don't think that this will work in all cases, because the support packages do not require any versioning in their title. `FromString` might complain that it's not a parse-able version.

We probably shouldn't attempt to find a version from the support package's name. It requires us to make too many assumptions about the compatibility, and I don't think that most of our proxies would necessarily benefit from being forced to follow these assumptions. Why would we say the acceptable range is `^4.0.0` as opposed to exactly equal, or `~4.0.0`? We would have to define that somewhere, and provide hints as to why a module is not passing checks for loading proxies so that developers could debug that. 

We should be able to untangle the cases without having to hack around by modifying it as such:

- Apply `ModuleVersion()` value if present, else apply `^0.0.0` instead of trying to parse the package name

This should have the effect of our return logic iterating through the choices correctly. 

Case 1: There are multiple `ModuleVersion` entries for our proxy:
```
proxy_details = { 'proxy_one': {'ModuleVersion': SemanticVersionRange(~4.0.0)}, 'proxy_two': {'ModuleVersion': SemanticVersionRange(~3.0.0)} }

tile_version = 4.0.0
```
Then, later in this function, you iterate through the proxies and "best match" the tile_version.

Case 2: There are multiple proxies, but each of them have the same version (i.e. no version set, but someone installed two):
```
proxy_details = { 'proxy_one': {'ModuleVersion': SemanticVersionRange(0.0.0)}, 'proxy_two': {'ModuleVersion': SemanticVersionRange(0.0.0)} }
tile_version = 2.0
```
How do we do it here? The algorithm needs to properly identify that we should still pick one of the two proxies even though the tile version is larger, right? Do we need a special case to handle this (can we just return early once this is noticed)? Should we log any debug info here so someone can turn up logs when their proxy doesn't load when they expect it to?

iotilecore/iotile/core/hw/hwmanager.py
mattrunchey: I'm not sure that the "latest match" is necessarily the correct thing here, though. We may need to define what "best match" is -- essentially, how do we decide what to use when multiple compatible proxies are available? Should we choose the first one, or the one that's got the biggest version?

anwu: Regarding your first comment... `latest_version` is checked in the first `for` loop, where it is updated as each proxy is visited. Not only that, but it marks itself as the latest version only when that proxy is compatible. That is why in the second `for` loop, we return right away when we see it because we've already iterated through the entire list of proxies.

anwu: Regarding your second comment... In this case, "best match" is the "latest version" AND "compatible".

iotilecore/iotile/core/hw/hwmanager.py
```diff
@@ -785,7 +788,47 @@ def get_proxy(self, short_name):
         if short_name not in self._name_map:
             return None
 
-        return self._name_map[short_name][0]
+        proxy_match = self.find_correct_proxy_version(self._name_map[short_name], version)
+        print("proxy_match: ", proxy_match)
+        return proxy_match if proxy_match is not None else self._name_map[short_name][0] 
+
+    def find_correct_proxy_version(self, proxies, version):
```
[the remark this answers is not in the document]
anwu: Understood.

iotilecore/iotile/core/hw/hwmanager.py
anwu: Sure, I can check that easily with a `isinstance(ModuleVersion, SemanticVersionRange)`. Curious, what would I do if it wasn't? Crap out and raise an error? Technically it would only happen in the `try` block, maybe it could just redirect to the except block i have for `AttributeError`.

iotilecore/iotile/core/hw/hwmanager.py
```diff
@@ -788,51 +788,47 @@ def get_proxy(self, short_name, version):
...
-                return proxy
-
-        # try matching based off of support package name
+                # If proxy does not have ModuleVersion(), use the name of the support package
+                stripped_proxy_version = str(proxy).split('.')[0].split('_')[-1]
+                module_version = SemanticVersionRange.FromString('^' + stripped_proxy_version + '.0.0')
+                proxy_details[proxy]['ModuleVersion'] = module_version
+
+            # Keep track of each proxy's minimum version required
+            proxy_details[proxy]['MinVersion'] = module_version._disjuncts[0][0][0]
```
mattrunchey: Don't you need to reset this above the loop instead? I think this causes you to always set the latest_version to the most recently checked proxy (i.e. we'll always return the *last* compatible proxy, even if it's not the actual biggest version)

mattrunchey: Nevermind, misread

iotilecore/iotile/core/hw/hwmanager.py
```diff
@@ -97,8 +97,9 @@ def _setup_proxies(self):
         for name, obj in proxy_classes:
             proxy_key = obj.__name__ + ':' + name
 
-            if proxy_key in self._proxies:
-                continue
+            # awu_10/01/19 - we want to add all proxies even if duplicate but diff version
+            # if proxy_key in self._proxies:
+            #     continue
 
             self._proxies[proxy_key] = obj
```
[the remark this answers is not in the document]
anwu: It's appended the the `self._name_map` a few lines down. We don't actually look at the `self._proxies` in our case if that's what you're wondering. We look at `self._name_map`.

iotilecore/iotile/core/hw/hwmanager.py
```diff
@@ -167,13 +168,15 @@ def get(self, address, basic=False, force=None, uuid=None):
             return tile
 
         name = tile.tile_name()
-        _version = tile.tile_version()
+        version = tile.tile_version()
```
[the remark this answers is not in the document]
anwu: I can't be 100% certain, since I'm not familiar with every single tile. I assume they all, even emulated ones, know that RPC as the status rpc, and will return that information... I can add a check I guess. I'm not too sure what to do if it doesn't return a version. Maybe it will just select "best match" somehow? Maybe the latest proxy?

iotilecore/iotile/core/hw/hwmanager.py
anwu: Agreed. Using support package name is a fickle method. As @mattrunchey and I discussed, we should try to seek a best match if `ModuleVersion` does return something useful. If it doesn't then we would pretty much have to select a proxy like we did before -- just choose one.

x24ling:
Don't forget to update `version.py` and `RELEASE.md` for iotile-core like you did for iotile-emulate.

iotilecore/RELEASE.md
```diff
@@ -2,8 +2,9 @@
 
 All major changes in each released version of `iotile-core` are listed here.
 
-## HEAD
+## 5.0.6
```
mattrunchey: 5.0.6 is already released -- you might need to rebase this branch with master https://pypi.org/project/iotile-core/#history

anwu: Thanks for that. Just rebased.

mattrunchey: Sorry, I commented a touch too soon. I was just working on merging Dima's changes to master, so you may have to do it one more time. It shouldn't touch any of the same files, at least, because I coudn't update RELEASE.md with the notes that it should actually have..

iotilecore/iotile/core/hw/hwmanager.py
```diff
@@ -785,7 +786,52 @@ def get_proxy(self, short_name):
...
+            except AttributeError:
+                # If proxy does not have ModuleVersion(), use None
+                module_version = None
+                least_version = SemanticVersion(0, 0, 0)
+
+            if module_version is None:
+                # Regardless if version is compatible, choose a best proxy for now
+                if min_version == SemanticVersion(0, 0, 0):
+                    best_proxy = proxy
+                    self.logger.debug("Using default proxy: {0}".format(proxy))
```
mattrunchey: I might rephrase this to say "Found a proxy without ModuleVersion info". The main culprit word here is "using" -- since we aren't necessarily using it, just found one

mattrunchey:
Overall comment -- I can't tell if we've got test coverage for the new behavior, particularly to check that we're returning the expected proxy for common cases, such as (but not exhaustively)

(1) Two proxies installed with different `SemVerRange`, one matching the tile `SemVer`
(1a) Above, but both match the tile, with one being a bigger number
(1b) Above, but both have the same `SemVerRange` starting version (i.e. =3.0.0 and ^3.0.0), we pick the first one it comes across
(1c) Above, but neither matches the `SemVer`. We should return `None`

(2) Two proxies, but only one has `SemVerRange`. 
(2a) Tile has SemVer that matches SemVerRange, return versioned proxy
(2b) Tile has SemVer that doesn't match, return versionless proxy

(3) Two proxies, neither have SemVerRange
(3a) Tile has SemVer, return the first one

anwu:
> (1) Two proxies installed with different `SemVerRange`
> (1a) both proxies match the tile, return the one with a bigger starting range
> (1b) neither matches the `SemVer`. We should return `None`

(1a)
Tile Version: `4.1.5`
Installed proxies:
iotile-support-con-nrf52832-3: ModuleVersion = None
iotile-support-con-nrf52832-4: ModuleVersion = ^4.0.0
iotile-support-con-nrf52832-5: ModuleVersion = ^4.1.5

In this case, both `iotile-support-con-nrf52832-4` and `iotile-support-con-nrf52832-5` match my tile of `4.1.5`. The proxy matching successfully chose `iotile-support-con-nrf52832-5` since it had the newer base version.

(1b)
Returns None, but the function call above that `get_proxy` will choose the first proxy it sees.

> (2) Two proxies that have the same `SemVerRange` starting version (i.e. =3.0.0 and ^3.0.0)
> (2a) we return the first proxy that gets checked

(2a)
Tile Version: `4.0.0`
Installed proxies:
iotile-support-con-nrf52832-3: ModuleVersion = None
iotile-support-con-nrf52832-4: ModuleVersion = ^4.0.0
iotile-support-con-nrf52832-5: ModuleVersion = =4.0.0

Proxy matching successfully returned `iotile-support-con-nrf52832-4`.

(2b)
Made my own extra test case. Same as (2a), but with tile version at `4.1.5`, proxy matching correctly chose `iotile-support-con-nrf52832-4`.

> (3) Two proxies, but only one has `SemVerRange`.
> (3a) Tile has `SemVer` that matches `SemVerRange`, return versioned proxy
> (3b) Tile has `SemVer` that doesn't match, return versionless proxy

(3a)
Tile Version: `4.0.0`
Installed proxies:
iotile-support-con-nrf52832-3: ModuleVersion = None
iotile-support-con-nrf52832-4: ModuleVersion = ^4.0.0

Proxy matching successfully returned `iotile-support-con-nrf52832-4`.

(3b)
Tile Version: `5.0.0`
Installed proxies:
iotile-support-con-nrf52832-3: ModuleVersion = None
iotile-support-con-nrf52832-4: ModuleVersion = ^4.0.0

Proxy matching selected `iotile-support-con-nrf52832-3` as expected.

> (4) Two proxies, neither have `SemVerRange`
> (4a) Tile has `SemVer`, return the first one

(4a)
Tile Version: `5.0.0`
Installed proxies:
iotile-support-con-nrf52832-3: ModuleVersion = None
iotile-support-con-nrf52832-4: ModuleVersion = None

Proxy matching actually returned `iotile-support-con-nrf52832-4. Based on the logic, proxy matching will actually select the LAST proxy without a `ModuleVersion` if none of the proxies have that function. The code will keep falling into the conditional `if min_version == SemanticVersion(0, 0, 0)`, thus it will keep updating the `best_proxy`.

iotilecore/iotile/core/hw/hwmanager.py
anwu: Gotcha. Fixed.

mattrunchey:
I should clarify -- I think that those tests should be implemented in a file like `test_proxy_matching.py` so that we make sure someone doesn't unintentionally break it later down the line. I may have been incorrect about the logic of which one to pick when there are multiple eligible (and I'm not suggesting we change it).

x24ling:
Test runs:
1. Use POD-1C 0xADC with version 3.2.4 with the following proxies:
```
iotile-support-con-nrf52832-2        2.15.4
iotile-support-con-nrf52832-3        3.2.4
iotile-support-con-nrf52832-4        4.0.0.dev2
```
Output
```
PS C:\Users\chris\iotile\con_btc1\con_nrf52832> iotile hw connect 0xadc controller quit            tile_name:  NRF52
tile_version:  (3, 2, 4)
proxy_match:  <class 'iotile_support_con_nrf52832_4.nrf52832_controller.NRF52832Controller'>
```

None of the above proxies have the function ModuleVersion(), iotile-core correctly handles this case and selects con-nrf-4

2. Use POD-1C 0xADC with version 3.2.4 with the following proxies:
```
iotile-support-con-nrf52832-2        2.15.4
iotile-support-con-nrf52832-3        3.2.4
iotile-support-con-nrf52832-4        4.0.0.dev9
```
Output
```
PS C:\Users\chris\iotile\con_btc1\con_nrf52832> iotile hw connect 0xadc controller quit
tile_name:  NRF52
tile_version:  (3, 2, 4)
proxy_match:  <class 'iotile_support_con_nrf52832_3.nrf52832_controller.NRF52832Controller'>
```
The con-nrf-4 proxy does have ModuleVersion(), iotile-core correctly determines that con-nrf-4 has a version number that is too high, so it selects con-nrf-3

3. Use POD-1C 0x820 with version 4.0.0 with the following proxies:
```
iotile-support-con-nrf52832-2        2.15.4
iotile-support-con-nrf52832-3        3.2.4
iotile-support-con-nrf52832-4        4.0.0.dev9
```
Output
```
(new_prod_workspace) PS C:\Users\chris\iotile\con_btc1\con_nrf52832> iotile hw connect 0x820 controller quit            tile_name:  NRF52
tile_version:  (4, 0, 0)
proxy_match:  <class 'iotile_support_con_nrf52832_4.nrf52832_controller.NRF52832Controller'>
```
iotile-core correctly selects con_nrf_4 proxy

anwu:
@mattrunchey I've added a test suite. Check out my most recent commit (632f1ae). Hopefully I designed it correctly/efficiently. Open to feedback.

anwu:
@mattrunchey I believe the tests are failing because this PR is dependent on a `lib_controller` PR. There were 3 test suites I modified in here to use the new `ReferenceControllerProxy`. Those were expected to fail without the new `lib_controller` support package.

PR #890, opened 2019-10-27
Add ability to start a periodic coroutine in the SharedLoop, and test

mattrunchey:
## Overview

So far, we've been scheduling our periodic coroutines by using while loops, such as:
```python
async def _run_analytics_report(interval=5):
  while 1:
    compute_metrics()
    await asyncio.sleep(interval)
```

This works for many cases where exactness of an interval isn't a critical factor, and when the functional logic time is insignificant compared to the interval. In production, you'd probably want a bit more sugar to handle exceptions (or you'll exit the function) and interact with your logging utilities.

However, you may want to decouple the repeat interval from the computation, and/or not have drift (though we are always subject to jitter). 

## Major Changes

There is a new ability to pass a coroutine in to a function that you want to run on a periodic basis. 

```python
    def launch_periodic_coroutine(self, cor, interval, *args, handle_exceptions=True, **kwargs):
```

You need to send it a coroutine (just the function, not the actual instantiated object -- i.e. send it `_run_analytics_report` not `_run_analytics_report()`)

The complete invokation might be:
```python
from iotile.core.utilities.async_tools.event_loop import SharedLoop

SharedLoop.launch_periodic_coroutine(_run_analytics_report, 300)
```

This repeatable can be cancelled at any time in the future by calling `.cancel()` on the `Future` or `asyncio.Task` that gets returned to you (depending on whether you're inside or outside of the event loop). 

If you want the repeatable to exit if an exception happens, you should send in the `handle_exceptions` as `False`. This is useful during development in particular while ironing out the repeatable function.

This function is drift-corrected, but due to the nature of when tasks get scheduled, might have start time and end time jitter.

## Additional Notes

An example use case might be if we want to track the number of devices we see in our list of visible/scannable PODs, and send an alert any time that number drifts below a set value. For example, in your script you may define a periodic function such as:

(note that you must have this code branch/PR pulled to run this at the time of writing -- it's not in master or any release)
```python
from iotile.core.hw import HardwareManager
from iotile.core.utilities.async_tools.event_loop import SharedLoop

import time

global pod_count  # don't do this - globals are bad, just a demo
pod_count = 0

async def _check_POD_count(alert_level=5):
  if pod_count < alert_level:
    print("We are seeing fewer pods than we expect. Only {} scannable, we expect at least {}.".format(pod_count, alert_level))
    # Send an email or slack message? Post to MQTT?
  else:
    print("we see {} PODs".format(pod_count))

def main():
  global pod_count  # don't do this - globals are bad, just a demo
  with HardwareManager(port='bled112') as hw:
    hw.scan()
    time.sleep(1)  # Let things settle a bit
    pod_count = len(hw.scan())

    SharedLoop.launch_periodic_coroutine(_check_POD_count, 5)
    while 1:
      pod_count = len(hw.scan())
      time.sleep(1) . # For demonstration purposes, scan more often than we alert check for.

main()

```

### DIFF
--- a/iotilecore/iotile/core/utilities/async_tools/event_loop.py
+++ b/iotilecore/iotile/core/utilities/async_tools/event_loop.py
@@ -617,6 +617,52 @@ class BackgroundEventLoop:
         else:
             self.loop.call_soon_threadsafe(_run_and_log)
 
+    def launch_periodic_coroutine(self, cor, interval, *args, handle_exceptions=True, **kwargs):
+        """Start a coroutine task that's meant to occur every n seconds
+
+        If this method is called from inside the event loop, it will return an
+        awaitable object.  If it is called from outside the event loop it will
+        return an concurrent Future object.
+
+        The coroutine will be scheduled in this function, but has no
+        guaranteed execution time.
+
+        Drift is corrected for, to some extent, so your function should execute
+        at the interval regardless of its execution time.
+
+        You can call `cancel()` on the object to stop it.
+
+        Args:
+            cor (coroutine): The coroutine that we wish to periodically run in the
+                background.
+            interval (int): How many seconds you want the calls to happen apart
+            handle_exceptions (bool): Whether you want the repeatable to exit or
+                continue if an exception is seen. Defaults to True.
+
+        Returns:
+            Future or asyncio.Task: A future representing the coroutine.
+
+            If this method is called from within the background loop
+            then an awaitable asyncio.Tasks is returned.  Otherwise,
+            a concurrent Future object is returned.
+
+            You should NOT await on the returned object unless you are
+            intending to cancel it from another place (high risk of locking)
+        """
+
+        if self.stopping:
+            raise LoopStoppingError("Could not launch coroutine because loop is shutting down: %s" % cor)
+
+        # Ensure the loop exists and is started
+        self.start()
+
+        cor = _repeat(cor, interval, self.loop, self._logger, handle_exceptions, *args, **kwargs)
+
+        if self.inside_loop():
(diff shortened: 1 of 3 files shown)


iotilecore/iotile/core/utilities/async_tools/event_loop.py
```diff
@@ -692,6 +735,25 @@ def _log_future_exception(future, logger):
...
+    """Run a coroutine every interval seconds.
+
+    Note that if your function doesn't finish in time, it will run again immediately
+    after the previous iteration is finished (it doesn't guarantee runtime).
+
+    Each repeatable will have its own tick generation to account for its own drift.
+    """
+    tg = _tick_generator(interval, loop)
+    while 1:
+        await asyncio.gather(cor(*args, **kwargs), asyncio.sleep(next(tg)),)
```
timburke: If `cor` raises an exception, the repeating loop will be stopped and no more invocations will happen.  Is that the desired behavior?  I could see it both ways but likely I think the expectation would be that the repeat loop keeps going even if one particular invocation raised an exception.

timburke: For observability, I would recommend logging the exception as well.

mattrunchey: Good point, yes. I amended the function signature to accept the argument `handle_exceptions`, defaulting to True. The behavior now is that unexpected exceptions will be logged, and unless the exception was an `asyncio.CancelledError` (which would be how a user manually cancels their loop). If you have a case where you want the exception to break out of the loop (for example, during application development), you'd just set the handler to false.

PR #901, opened 2019-11-15
Added optional awaitable to return a "pointer" to the device

timothyfroehlich:
This script allows arch-node-tasks (and presumably others) to launch a virtual device, but doesn't return any references to the virtual device it creates, preventing the calling task from accessing it's info (such as any config values it's loaded in). This PR adds an optional future that will be set to the virtual device once it's initialized. Since the script launches coroutines then sleeps forever, asyncio futures seemed the best (only?) to get the reference back for something to use.

There is no functionality change to the script if the new argument isn't specified.

### DIFF
--- a/iotilecore/iotile/core/scripts/virtualdev_script.py
+++ b/iotilecore/iotile/core/scripts/virtualdev_script.py
@@ -47,7 +47,13 @@ def configure_logging(verbose):
 
 
 def main(argv=None, loop=SharedLoop, device_f=None):
-    """Serve access to a virtual IOTile device using a virtual iotile interface."""
+    """Serve access to a virtual IOTile device using a virtual iotile interface.
+
+    Args:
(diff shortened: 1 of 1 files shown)


iotilecore/iotile/core/scripts/virtualdev_script.py
```diff
@@ -134,8 +140,10 @@ def main(argv=None, loop=SharedLoop):
             raise
 
         started = True
-
         print("Starting to serve virtual IOTile device")
+        if device_f:
```
mattrunchey: I think for safety you might also want to check the device existing. Technically we should be raising if the instantiation fails, but it's easy to check regardless:
```suggestion
        if device_f and device is not None:
```

timothyfroehlich:
Will be implementing the second solution. Assuming it works fine, I will retract this PR.

timothyfroehlich:
Changes no longer needed, better solution found.

Issue #893, opened 2019-11-07
Intermittent test failures in `iotile_transport_bled112`


mattrunchey:
In Azure pipelines, sometimes the OSX test environments seem to occasionally straight up hang on `iotile_transport_bled112`. I've seen it enough now that we might want to consider investigating this, as it's quite annoying to just rerun checks due to a timeout (it's still better than risking a buggy merge). 

```
  Running setup.py develop for iotile-ext-cloud
Successfully installed iotile-cloud-0.9.12 iotile-ext-cloud
Testing iotile_ext_cloud: PASSED (20.7 seconds)
Testing iotile_transport_awsiot: PASSED (1.3 seconds)
Testing iotile_transport_bled112: 
##[error]The operation was canceled.
```

A few reference links to the builds: 

https://dev.azure.com/iotile/coretools/_build/results?buildId=599
https://dev.azure.com/iotile/coretools/_build/results?buildId=656

Unfortunately, we don't see any test logs because Azure kills the container after 60 minutes.

I'm going to attempt to get a recreation on my machine, hopefully I can get us started on the actual test that's failing with more test output...

mattrunchey:
I can't get these tests to fail so far on my machine when using the same base command (`pytest test -s` from within the `bled112` folder).

The only additional output I'm getting is some test cleanup failures. I'm going to see if I can't clean that up, but I doubt it's the reason we're hanging or it'd be more consistent.

```
Error running finalizer for task LegacyAdapter (BLED112Adapter)
Traceback (most recent call last):
  File "/Users/mrunchey/gitrepos/coretools/iotilecore/iotile/core/utilities/async_tools/event_loop.py", line 193, in stop
    await result
  File "/Users/mrunchey/gitrepos/coretools/iotilecore/iotile/core/hw/transport/adapter/async_wrapper.py", line 103, in stop
    await self._execute(self._adapter.stop_sync)
  File "/usr/local/Cellar/python35/3.5.6_2/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/futures.py", line 381, in __iter__
    yield self  # This tells Task to wait for completion.
  File "/usr/local/Cellar/python35/3.5.6_2/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/tasks.py", line 310, in _wakeup
    future.result()
[...]

PR #906, opened 2019-11-25
Update Pylint config

Watkurem:
## Overview & Major Changes
- Updates the pylint config to match the template used by newer pylint versions (to ease diffing with default config, for example);
- Prepares the pylint config of coretools to be used as global default for all Arch projects.
Related issue: https://github.com/iotile/archfx_cloud/issues/251
Related PR: https://github.com/iotile/archfx_cloud/pull/373
(No need to read through all discussion there, but may provide some context.)

### Additional Notes
In case total PR would be confusing (since sections moved around, many style changes), individual commits show atomic, readable changes.
All commits until `2cdf9bd`: only style updates.
`3378d64` to `9972322`: new (from newer pylint versions) defaults.
`c4505db` onwards: non-default changes, each one explained in commit message.
Can be squashed after review if that's the preferred approach.

#### Unsolved challenge
Since subprojects inside this repo are not Python packages, the `.pylintrc` is not automatically applied when current directory is not top-level.
Example:
`$ pylint iotilecore/iotile/core/utilities/async_tools`
PWD is `<coretools>` (project root). This command will use the top-level `.pylintrc`.
`$ pylint async_tools`
PWD is `<coretools>/iotilecore/iotile/core/utilities`. This command will NOT use the top-level `.pylintrc`, because Pylint looks for an rcfile in parent directories only as long as each of them is a package. `iotilecore` directory is not a package (has no `__init__.py`) and search stops there, at which point Pylint defaults to a built-in config.

Ways to deal with it:
1. Specify rcfile directly (either via `--rcfile` argument or via `PYLINTRC` env variable). Somewhat cumbersome. If using an editor with Pylint support, depends on editor's ability to run different command/set of env variables per project;
2. Add a symlink to the main rcfile to each subproject. Will work fine, but depends on contributor's OS/FS to support symlinks. Windows with NTFS should be able to handle it, OS X/Linux with most filesystems, obviously, too;
3. Add an `__init__.py` file to each subproject. Will work fine, should be OS/FS-independent. Seems like a best approach to me, but maybe breaks some Python packaging (`setup.py`/`setup.cfg`) logic?
4. Pylint should soon add an `rcfile` option to the pylintrc. This will allow to 'include' other rcfiles, and so we could have a `.pylintrc` in each subproject that would have a single `rcfile=../.pylintrc` line in it. Depends on Pylint upstream and on contributors using the newest Pylint version, though.

Should I go for 2 or 3, or do we proceed with 1 with no further changes?


Watkurem:
I'm not sure why the test failed, but at first glance it does not seem related to pylint at all.

mattrunchey:
> I'm not sure why the test failed, but at first glance it does not seem related to pylint at all.

Sorry, this was me. I've been investigating occasional test timeouts but haven't resolved it yet (see https://github.com/iotile/coretools/issues/893 if you're curious). I re-ran the test to get it to pass but sometimes the hooks don't line back up with the PR. 

Particularly since this PR is about linting, we can disregard test failures and get it all in at once.

mattrunchey:
Overall, I think this makes the pylint file look much nicer and easier to handle. Thanks for cleaning that up!

All of the updated defaults look fine to me at first glance. Nothing stood out to me as strange there.

> Enable len-as-condition, redefined-builtin, redefined-outer-name

In the last commit, you re-enabled a few things that got disabled in history. I'm not sure why they were disabled in the first place, but there are some potential reasons for it (@timburke may remember the original intention for them though).

- `len-as-condition` (this is the same as `C1801`) seems to be a bit controversial, and even wrong when you are using numpy (https://stackoverflow.com/a/43857007/10377552), which may be why it was disabled. It seems like this was added over a weird PEP8 guideline about the pythonic way to check list size. I would be ok with actually leaving this one disabled from looking at those convos
- `redefined-outer-name` (`W0621`) was probably disabled due to test code, but it actually looks like the PyCQA people have a way to enable it without catching fixtures based on issue https://github.com/PyCQA/pylint/issues/1535 . Perhaps that could be a way to enable it without false flagging fixtures?
- `redefined-builtin` I'm not sure what this one was addressing, but perhaps it was our use of an `exit` override somewhere (see https://github.com/PyCQA/pylint/issues/2458). It doesn't seem like that would be a prevalent enough reason to fully disable the warning, though. Maybe @timburke can shed some light on this reason this was disabled.

Anyway. I think these changes are a good starting point. We should just double check that we re-enable the three checks above without missing the reason they were added in the first place.

I'm not sure what the `setup.cfg` is meant for. You only added it to a few of the modules in this project (there are several in the `transport_plugins/*` directories). Is this meant for a particular editor, or a particular project? I don't see these as being applicable to everyone's development environment, or contributing to a CI desire, but perhaps I just need you to clarify that a bit.

> Since subprojects inside this repo are not Python packages, the .pylintrc is not automatically applied when current directory is not top-level.


I think it might help to separate two things here:

1. Linting as you write code (i.e. usually you only lint the file you're looking at in the editor)
2. Linting as a CI blocker or check

The first one shouldn't have any relation to where a RC file lives, because I can't think of an editor that doesn't just let you configure and run `pylint --rcfile=ENVVAR {mycurrentfile}` in the background, and then highlighting up your code with the results, and the second best thing is to just use your terminal to run a copy pasted command just like that.

The second one is solvable by writing a customized command that runs pylint on each project separately much in the way that we build and test each module's tests separately. We have a special script for handling this case with pytest as well for coretools, in `scripts/test.py` (https://github.com/iotile/coretools/blob/master/scripts/test.py). We also do not need to solve this problem now, because we're still far away from adding that to the review process. Other projects that have simpler pytest or other running routines can pretty easily solve this by writing a one-time script in the test configuration or steps file.

Watkurem:
>`len-as-condition` (this is the same as C1801) seems to be a bit controversial, and even wrong when you are using numpy (https://stackoverflow.com/a/43857007/10377552)

From what's referenced in that SO thread and my own experimentation just now, it seems like "pylint-approved" way of handling this is by using `if len(x) == 0:` instead of `if len(x):`.
I personally feel like "Empty means False" is as Pythonic as e.g. 4 space indent, and an explicit way to deal with it like that would work great, while enforcing PEP8 approach where it's not required.
Still agree that it's ok to keep it disabled if that's the preferred way.
>but it actually looks like the PyCQA people have a way to enable it without catching fixtures based on issue

Not really. What that issue is talking about (as actually implemented in linked PR) is that the `redefined-outer-name` check ignores the values specified in `ignored-argument-names`. So, for it to work with fixtures, you would have to name all fixtures consistently (e.g. all with a `fixture_` prefix), and add a regexp for that consistent name to the `ignored-argument-names`.
To me this does not seem like a big improvement over a module-level `#disable` for this check, one per test module.
>I'm not sure what the setup.cfg is meant for.

Unlike `pylint` (which also supports `setup.cfg`-based configuration, btw) the `pycodestyle` (former `pep8`) and `flake8` tools support _only_ the `setup.cfg` configuration. So for them to accept a different line length limit, we have to add this configuration.
`pep8` is recommended to always be used alongside `pylint` in our style guide. `flake8` is just what I use (it's an aggregate checker that includes `pycodestyle` - former `pep8`), so that section can be safely removed if it isn't useful.
>You only added it to a few of the modules in this project

This is caused by the way those tools look up their configuration. Pylint, for example, continues looking in alternative places if it does not find a `pylint` section in `setup.cfg`. So, since we do not have such section, it finds a `.pylintrc` (as long as it can traverse up packages).
`pycodestyle`/`pep8`/`flake8`, though, stop looking as soon as they find _any_ `setup.cfg`. If there is no section relevant to them, they just assume the defaults.
So I have added a top-level `setup.cfg` for them to find, but two subprojects had their own `setup.cfg`, so for those subprojects the top-level one did not apply - I had to copy the sections to each of them, too.

Watkurem:
It seems like for both `flake8` and `pycodestyle`, a `--config` argument is available, same as `--rcfile` for `pylint`.
If we choose to rely on explicit overriding instead of config auto-discovery, only the top-level `setup.cfg` is required.

mattrunchey:
Is there a reason that we're trying to support those other linters? I only think that I used `pycodestyle` because the line length check wasn't fully working in my editor, but now that it's fixed I think they are mostly redundant. Unless we're trying to also standardize their use and integrate them in to CI, it seems a touch out of place.

Watkurem:
> Unless we're trying to also standardize their use and integrate them in to CI, it seems a touch out of place.

If I looked in a wrong place and got confused about recommended tools, please just confirm it and I'll revert the `setup.cfg` additions.

mattrunchey:
I vote that the `setup.cfg` be removed, because I don't expect most people to actually end up using them (and instead use their editor-wide settings).  The spirit of this `.pylintrc` change is to finally set an updated style foundation, and less about providing robust editor configuration (I consider that a process improvement more than something that can be addressed by having a lot of config files)

mattrunchey:
@dkarchmer did you get a chance to glance at the commits/discussion here to weigh in?

Watkurem:
I've reverted the `setup.cfg` changes for other linters

PR #910, opened 2019-12-03
Preliminary work to Port BLED112 device adapter fully to asyncio

timburke:
### Overview

This PR is into a long-running branch `feat-bled112-refactor`.  It is the first in a series of PRs that are designed to move bled112 completely over to asyncio.  This PR just contains preliminary enhancements and fixes to `iotile-core` and testing utilities in `iotile-transport-bled112` needed for the port to happen.

### Key Changes

1. Fixes the behavior of `AsynchronousModernWrapper`  to not call `stop()` twice when invoked by a HardwareManager that has `close()` called on it explicitly.
2. Fixes the very old behavior of our pyserial mocking wrapper in bled112's test files to use a proper queue for simulated reads so that we can support `cancel_read()` which is a pyserial behavior that lets us cancel a read that is in progress.  This will be used to decrease the CPU usage of running a `virtual_device` significantly.
3. Updates `OperationManager` to be ready for primetime use inside of the new bled112 code.  For now these are just a few tweaks to the `OperationManager`.  The next PR will use them to great effect.
4. Fixes a pug in python 3.8.0 that causes an exception to be thrown when closing down the BackgroundEventLoop.  This is going to be fixed in python 3.8.1 anyway (it's a CPython regression) and only affects windows.  This code adds a hook to check and do a point fix if we happen to be running on python 3.8.0 and Windows.

### OperationManager

The primary purpose of this PR is to get `OperationManager` ready for use in the bled112 refactor.  I'm splitting the actual refactor into a separate PR since it is a larger change.  `OperationManager` is a new concept designed to simplify the writing of complex, asynchronous networking or communications code.  That kind of code has a pattern where you trigger an action and then need to wait for asynchronous events from your remote partner telling you what happened.  Often you need to then respond to those events, causing you to wait for new events.  

Each time you respond and wait for a new event, there's typically a small race condition where you install the wait handler after you write the command that triggers the new event.  So, if the event comes too fast, you get it before your handler is installed to listen for it.  

There are two traditional fixes for this:

1. Use a single thread to synchronize all packets received from the remote partner and only write and read from that thread.  This way there are no races because you queue packets and only ever read them from a single thread.  The downside is that this paradigm doesn't let you have multiple parallel operations going on because there's only a single thread of control that can access the packet stream.
2. Use complex callback code.  This lets you write "asynchronous" operations and have multiple running at the same time but it imposes a significant maintainability tax since "simple" operations turn into complex chains of callbacks.

OperationManager is designed to let you write your operations as coroutines and have them run in parallel with a single thread dispatching events to each coroutine and waking them up as the next step in their state machine is triggered.  It uses cooperative `pause`/`unpause` calls to know when its safe to dispatch new packets because no one is in the process of setting up a new wait handler.

### DIFF
--- a/iotilecore/iotile/core/utilities/async_tools/event_loop.py
+++ b/iotilecore/iotile/core/utilities/async_tools/event_loop.py
@@ -16,6 +16,8 @@ register then as subtasks and their parent task is in charge of stopping
 them cleanly in the correct order.
 """
 
+import sys
+import platform
 import time
 import asyncio
 import inspect
@@ -280,6 +282,8 @@ class BackgroundEventLoop:
         self._logger = logging.getLogger(__name__)
         self._loop_check = threading.local()
         self._pool = None
+
+        _check_patch_python_3_8_0()
 
     def start(self, aug='EventLoopThread'):
         """Ensure the background loop is running.
@@ -765,12 +769,37 @@ async def _repeat(cor, interval, loop, logger, handle_exceptions, *args, **kwarg
             logger.exception("Exception in repeating coroutine; exiting repetition: %s", cor, exc_info=True)
             raise
 
+
 def _tick_generator(period, loop):
     t1 = loop.time()
     count = 0
     while True:
         count += 1
         yield max(t1 + count * period - loop.time(), 0)
+
+
+def _check_patch_python_3_8_0():
+    """Python 3.8 on windows switches event loops but has a bug.
+
+    The event loop changes to ProactorEventLoop from SelectorEventLoop which
+    is generally good but in python 3.8.0 it has a bug that causes an
+    exception to be raised at exit time since it tries to remove a
+    KeyboardInterrupt handler when running in a background thread.  This
+    behavior is fixed in python 3.8.1.
+    """
+
+    if platform.system() != "Windows":
+        return
+
+    if sys.version_info < (3, 8, 0) or sys.version_info >= (3, 8, 1):
+        return
+
+    logger = logging.getLogger(__name__)
+
+    logger.info("Patching event loop policy to work around buggy ProactorEventLoop on python 3.8.0 (version_info=%s)",
+                sys.version_info)
+    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
+
 
 # Create a single global event loop that anyone can add tasks to.
 SharedLoop = BackgroundEventLoop()  # pylint:disable=invalid-name;This is for backwards compatibility.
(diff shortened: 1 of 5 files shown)


iotilecore/iotile/core/utilities/async_tools/event_loop.py
```diff
@@ -765,13 +769,38 @@ def _log_future_exception(future, logger):
...
 def _tick_generator(period, loop):
     t1 = loop.time()
     count = 0
     while True:
         count += 1
         yield max(t1 + count * period - loop.time(), 0)
 
+
+def _check_patch_python_3_8_0():
+    """Python 3.8 on windows switches event loops but has a bug.
```
mattrunchey: Probably not necessary, but was there a link to the python bug/fix thread for posterity? I find it fun to read the gritty details sometimes :)

iotilecore/iotile/core/utilities/async_tools/operation_manager.py
```diff
@@ -254,11 +271,75 @@ def wait_for(self, timeout=None, **kwargs):
...
-        future = self._add_waiter(spec)
+        future = self._add_waiter(spec, pause=pause)
         future.add_done_callback(lambda x: self._remove_waiter(spec, future))
 
+        if unpause:
+            self.unpause()
+
         return asyncio.wait_for(future, timeout=timeout)
 
+    async def gather_until(self, gather_specs, until, timeout=None, *, pause=False, unpause=False):
```
mattrunchey: Looking at this function alone, I can't easily tell what the arguments coming in are supposed to be (`gather_specs` and `until`)

iotilecore/iotile/core/utilities/async_tools/operation_manager.py
```diff
@@ -316,8 +402,45 @@ def wait_for(self, timeout=None, **kwargs):
...
 
+        if len(processors) > 0 and wait:
+            results = await asyncio.gather(*processors, return_exceptions=True)
+            for proc, result in zip(processors, results):
+                if isinstance(result, Exception) and not isinstance(result, asyncio.CancelledError):
+                    self._logger.error("Error running processor %s: %s", proc, result)
+
         return not ignored
 
+    async def process_all(self):
```
mattrunchey: If this is a public method, should it have a short docstring for consistency (or is that overkill)? Saw it in a few places, and it seems obvious enough, but we're slightly diverging from our style guide I think (perhaps the language could be softened... particularly if we start introducing type hints)

iotilecore/iotile/core/utilities/async_tools/event_loop.py
timburke: Updated with a link: 
https://bugs.python.org/issue34679

iotilecore/iotile/core/utilities/async_tools/operation_manager.py
timburke: I updated all of the docstrings to make them compliant with our style guide.  Thanks for checking.

iotilecore/iotile/core/utilities/async_tools/operation_manager.py
timburke: Docstrings updated in latest commits.

PR #913, opened 2019-12-04
Port bled112 virtual device to use new asyncio api

timburke:
### Overview

This PR replaces the `BLED112CommandProcessor`, which is a background thread that manages a bled112 adapter with an asyncio version that does not need a background thread.  The `BLED112DeviceServer` that allows serving virtual devices is then ported to use the new asyncio based interface.

Additional basic tests are written for the asyncio interface and a new hardware loopback test is made to allow testing a bled112 client connecting to a virtual device.  This test just verifies connecting and streaming currently but will be enhanced in the future to support a full range of 'acceptance' style tests to make sure that all major operations work on the bled112 device adapter and server.

The tests can be run by passing `--hardware` to `pytest` while running the test suite.

This is mainly a behind the scenes PR that demonstrates it's possible to remove the background bled112 thread completely.  Since it was easier to port the virtual-device than the device adapter I started with that.  

The subsequent PRs will do some of the heavier lifting to port the device adapter itself.

Note the main class here: `AsyncBLED112` is intended to be internal to the bled112 package so it's not considered public.  Its a direct port of `BLED112CommandProcessor` to a coroutine based interface with some minimal cleanup.  The old command processor module is removed since it's no longer used.  (There is still a copy remaining that is used by the device adapter).


transport_plugins/bled112/iotile_transport_bled112/hardware/async_bled112.py
```diff
@@ -0,0 +1,459 @@
...
+
+                    value = result['data']
+                    assert len(value) == 2
+                    value, = struct.unpack("<H", value)
+
+                    last_char['client_configuration'] = {'handle': handle, 'value': value}
+
+        return {'services': services}
+
+    async def write_highspeed(self, conn, handle, data, progress_callback=None):
```
mattrunchey: I believe `handle` is supposed to be an `int`, based on how `_write_handle` is using it. You mentioned that this is mostly a port of an old class, so it probably didn't have a docstring to begin with.

timburke: It is.  I'm going to add the type annotation later here because I think this method will actually get removed from this class before the final merge of this branch as I see how the next phase is shaping up to look.

PR #915, opened 2019-12-06
Add support for device servers to indicate fatal errors.

timburke:
Currently, the virtual-device script has no way of knowing if the underlying device server that it is hosting has encountered a fatal error and is no longer running.  Since the server runs in a background event loop, exceptions don't naturally cause the main script to exit.

This commit adds a `failed` Future() on all DeviceServers that can be set to indicate that the server has encounted a fatal error condition and stopped operation.

The virtual-device script is updated to wait on this future rather than just sleeping until Ctrl-C is hit.

An example of the new usage is:

```
$ virtual-device bled112 simple
```

If you unplug the BLED112 dongle that is serving the virtual device, it can now cleanly exit the virtual_device program.  FYI, for this new functionality to work, it requires additional changes (not part of this PR) that actually set the `failed` future when there's a fatal serial port error.  This PR is just the baseline infrastructure changes to allow for that work to propagate all the way to the `virtual-device` script.

### DIFF
--- a/iotilecore/iotile/core/scripts/virtualdev_script.py
+++ b/iotilecore/iotile/core/scripts/virtualdev_script.py
@@ -2,15 +2,15 @@
 
 import argparse
 import sys
+import asyncio
 import logging
 import json
-import time
-from typedargs.doc_parser import ParsedDocstring
 from iotile.core.dev import ComponentRegistry
 from iotile.core.exceptions import ArgumentError
 from iotile.core.hw.virtual import BaseVirtualDevice
 from iotile.core.hw.transport import AbstractDeviceServer, StandardDeviceServer, VirtualDeviceAdapter
 from iotile.core.utilities import SharedLoop
+from typedargs.doc_parser import ParsedDocstring
 
 
 def one_line_desc(obj):
@@ -89,6 +89,8 @@ def main(argv=None, loop=SharedLoop):
 
     configure_logging(args.verbose)
 
+    logger = logging.getLogger(__name__)
+
     config = {}
     if args.config is not None:
         with open(args.config, "r") as conf_file:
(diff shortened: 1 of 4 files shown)


iotilecore/iotile/core/scripts/virtualdev_script.py
```diff
@@ -140,19 +143,24 @@ def main(argv=None, loop=SharedLoop):
         if stop_immediately:
             return 0
 
-        # We need to periodically process events that are queued up in the interface
+        # Make sure to check
```
mattrunchey: This comment looks like it may be missing a few words?

iotilecore/iotile/core/scripts/virtualdev_script.py
```diff
@@ -140,19 +143,24 @@ def main(argv=None, loop=SharedLoop):
...
         while True:
-            time.sleep(0.5)
+            loop.run_coroutine(supervise_server, server, timeout=0.2)
 
     except KeyboardInterrupt:
         print("Break received, cleanly exiting...")
+    except:
```
mattrunchey:
```suggestion
    except:  # pylint:disable=bare-except;We want to exit cleanly and log the exception
```

iotilecore/iotile/core/scripts/virtualdev_script.py
```diff
@@ -162,6 +170,21 @@ def main(argv=None, loop=SharedLoop):
     return 0
 
 
+async def supervise_server(server, timeout=0.5):
```
mattrunchey: Maybe not necessary, but what type of object does `server` expect? From contest I gather it is supposed to be `AbstractDeviceServer` and derived classes.

timburke: annotation added

Issue #919, opened 2019-12-10
Missing Crypto dependency for iotile.core


lelithium:
Steps to replicate:
```
python3 --version
Python 3.7.3

python3 -m venv venv
. venv/bin/activate
pip install iotile-core
python -c "import iotile.core.hw"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "[...]/venv/lib/python3.7/site-packages/iotile/core/hw/__init__.py", line 2, in <module>
    from .proxy import TileBusProxyObject
  File "[...]/venv/lib/python3.7/site-packages/iotile/core/hw/proxy/__init__.py", line 9, in <module>
    from .proxy import TileBusProxyObject
  File "[...]/venv/lib/python3.7/site-packages/iotile/core/hw/proxy/proxy.py", line 15, in <module>
    from ..virtual import pack_rpc_payload, unpack_rpc_payload
  File "[...]/venv/lib/python3.7/site-packages/iotile/core/hw/virtual/__init__.py", line 119, in <module>
    from .virtualdevice_base import BaseVirtualDevice, AbstractAsyncDeviceChannel
  File "[...]/venv/lib/python3.7/site-packages/iotile/core/hw/virtual/virtualdevice_base.py", line 9, in <module>
    from ..reports.individual_format import IndividualReadingReport
  File "[...]/venv/lib/python3.7/site-packages/iotile/core/hw/reports/__init__.py", line 3, in <module>
    from .signed_list_format import SignedListReport
  File "[...]/venv/lib/python3.7/site-packages/iotile/core/hw/reports/signed_list_format.py", line 10, in <module>
    from iotile.core.hw.auth.auth_provider import AuthProvider
  File "[...]/venv/lib/python3.7/site-packages/iotile/core/hw/auth/auth_provider.py", line 6, in <module>
    from Crypto.Cipher import AES
ModuleNotFoundError: No module named 'Crypto'
```

This differs from the expected behavior, stated [here](https://github.com/iotile/coretools#working-with-encrypted-device-data), that `pycrypto` is only required for working with encrypted data.

The issue was introduced during PR #879. It was not flagged in the tests due to `pycryptodome` being installed manually in the `tox.ini` file.

**Proposed fix**: Add `pycryptodome` to the `setup.py` file for `iotile-core`. `pycrypto` is unmaintained, and `pycryptodome` offers a drop-in replacement (see [this article](https://blog.sqreen.com/stop-using-pycrypto-use-pycryptodome/))

[...]

PR #920, opened 2019-12-11
Fix missing Crypto dependency #919

DimaWittmann:
We need to restore the old behavior where iotile-core only imported Crypto if it was asked to perform an operation that needs it.

In case ble broadcast enabled on the device and the library is missing on the host then logs with explanation will be generated.

The PR adds small improvement of bled112_auth error handling.


transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -699,14 +698,15 @@ def _parse_v2_advertisement(self, rssi, sender, data):
...
 
             nonce = generate_nonce(device_id, timestamp, reboot_low, reboot_high_packed, counter_packed)
 
             try:
-                decrypted_data = decrypt_payload(key, data[7:], nonce)
-            except ValueError:
+                from Crypto.Cipher import AES
+                decrypted_data = decrypt_payload(AES, key, data[7:], nonce)
+            except (ImportError, ValueError):
                 self._logger.warning("Advertisement packet is not verified", exc_info=True)
```
timburke: If left as is, this would log a warning for every encrypted broadcast packet received, which could be hundreds per second.

I would recommend refactoring this as:

```python
# At module level
try:
    from Crypto.Cipher import AES
    _HAS_CRYPTO = True
except import:
   _HAS_CRYTO = False


# inside DeviceAdapter __init__()
def __init__():
    if not _HAS_CRYPTO:
        self._logger.warning("pycryptodome is not installed, encrypted v2 broadcasts will be dropped.")


# inside _parse_v2_advertisement
def __parse_v2_advertisement():

    if is_encrypted and not _HAS_CRYTPO:
       return # Whatever return value indicates the broadcast was invalid
```

DimaWittmann: You are right, thank you for suggestion, I wasn't sure how to properly log error case

PR #921, opened 2019-12-12
Fix typo in LinebufferUI error suggestion

timothyfroehlich:
## Overview & Major Changes

When attempting to launch `iotile hw watch_broadcasts` without asciimatics installed, an error would be displayed that suggested installing `iotilecore[ui]` This wouldn't work, the proper package name is `iotile-core[ui]`


timburke:
Is anything blocking merging this?  It's been approved for a few days.

PR #943, opened 2020-01-31
Comment out "RPC in progress" check

DimaWittmann:
Recently a check was added to send_rpc_async
```Python
        if self.check_is_rpc_in_progress(found_handle, services):
            callback(conn_id, self.id, False, 'RPC still in progress', None, None)
            return
```
This check periodically causes RPCs to fail with exception:
```
E           iotile.core.hw.exceptions.DeviceAdapterError: DeviceAdapterError: Operation send_rpc on conn 0 failed because Error writing to handle
E           Additional Information:
E           reason: Error writing to handle
E           operation: send_rpc
E           conn_id: 0
```

The impelemtation needs to be revised.

### DIFF
--- a/transport_plugins/bled112/RELEASE.md
+++ b/transport_plugins/bled112/RELEASE.md
@@ -11,6 +11,7 @@ All major changes in each released version of the bled112 transport plugin are l
   generated the same way it is done for user key
 - Add the password-based authentication method
 - Update the authentication flow
+- Disable "RPC in progress" check
 
 ## 3.0.3
 
--- a/transport_plugins/bled112/iotile_transport_bled112/bled112.py
+++ b/transport_plugins/bled112/iotile_transport_bled112/bled112.py
@@ -311,10 +311,6 @@ class BLED112Adapter(DeviceAdapter):
 
         services = self._connections[found_handle]['services']
 
-        if self.check_is_rpc_in_progress(found_handle, services):
-            callback(conn_id, self.id, False, 'RPC still in progress', None, None)
-            return
-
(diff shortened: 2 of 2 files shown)


mattrunchey:
There were some super strange errors on Azure pipelines (like `gcc` download timing out). I reran them and it's all passing now.

timburke:
> This makes sense to me. We don't want to prohibit an asynchronous RPC from being started because an RPC is already in progress. Otherwise, that sort of defeats the purpose of it being asynchronous.

@anwu @DimaWittmann I just want to make sure the underlying architecture is clear.  There are **no asynchronous rpcs** that can be invoked via CoreTools.  

All RPCs throughout all iotile devices are the equivalent of synchronous function invocations like calling a `C` subroutine.  The caller calls the RPC and waits synchronously for it to finish.  This is the only kind of RPC that we have.  I want to make sure it's clear that from the callers point of view (over Bluetooth CoreTools `HardwareManager` is the caller), there is never an asynchronous rpc.  All RPCs synchronously run to completion before returning control back to `HardwareManager`.  

This synchronous RPC architecture is very important and a fundamental part of how IOTile devices are designed to work.  The check that was being done here is not related to whether an RPC is synchronous or asynchonous since all rpcs are synchronous at this level but rather whether a previous caller invoked a synchronous RPC and then stopped waiting for the response.  From the POV of the device, the RPC is still synchronously running but CoreTools does not know that.  So the question is, how can CoreTools discover if there is currently a synchronous RPC running over bluetooth if for whatever reason it's not sure.

Let me know what questions you have about this so I can make sure it's clear.  This is a key part of the architecture of how RPCs are designed that has a lot of implications so I want to make sure we're all on the same page about how it works.

anwu:
I understand now. Thanks for the clarification.

PR #945, opened 2020-02-12
Update decription logic for signed reports

DimaWittmann:
Recently two new auth methods were introduced: NullKey and PasswordBased. This change has broken the signed list encryption/decryption logic. 

The purpose of the PR is to simplify logic by introducing explicit report type and conversation function and replace hardcoded key type values with defined variables in AuthProvider.

### DIFF
--- a/iotilecore/iotile/core/hw/reports/signed_list_format.py
+++ b/iotilecore/iotile/core/hw/reports/signed_list_format.py
@@ -4,12 +4,18 @@ import datetime
 import struct
 import hashlib
 import hmac
+from enum import IntEnum
 from .report import IOTileReport, IOTileReading
 from iotile.core.utilities.packed import unpack
 from iotile.core.exceptions import NotFoundError, ExternalError
 from iotile.core.hw.auth.auth_provider import AuthProvider
 from iotile.core.hw.auth.auth_chain import ChainedAuthProvider
 
+class ReportSignatureFlags(IntEnum):
+    """Types of stream report signature"""
+    SIGNED_WITH_HASH = 0
+    SIGNED_WITH_USER_KEY = 1
+    SIGNED_WITH_DEVICE_KEY = 2
 
 class SignedListReport(IOTileReport):
     """A report that consists of a signed list of readings.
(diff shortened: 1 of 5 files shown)


iotilecore/iotile/core/hw/reports/signed_list_format.py
```diff
@@ -20,9 +21,34 @@ class SignedListReport(IOTileReport):
 
     ReportType = 1
 
+    class StreamType(Enum):
+        """Types of stream reports"""
+        INDIVIDUAL_READING = 0
+        SIGNED_WITH_HESH = 1
```
mattrunchey: Is this meant to be HASH?

iotilecore/iotile/core/hw/reports/signed_list_format.py
```diff
@@ -20,9 +21,34 @@ class SignedListReport(IOTileReport):
...
+        """Converts the stream type into the key type that was used to encryped this stream report
+
+        Args:
+            StreamType: a type of the stream report
+
+        Returns:
+            int: type of key, see AuthProvider
+        """
+
+        if stream_type in [cls.StreamType.INDIVIDUAL_READING, cls.StreamType.SIGNED_WITH_HESH]:
```
mattrunchey: Same comment -- HASH not HESH

mattrunchey:
Looks like something that changed here broke a test though

iotilecore/iotile/core/hw/reports/signed_list_format.py
DimaWittmann: Yes, thank you

iotilecore/iotile/core/hw/reports/signed_list_format.py
```diff
@@ -23,6 +29,41 @@ class SignedListReport(IOTileReport):
...
+        """Converts the stream type into the key type that was used to encryped this stream report
+
+        Args:
+            ReportSignatureFlags: a type of the stream report
+
+        Returns:
+            int: type of key, see AuthProvider
+        """
+
+        if stream_type in [ReportSignatureFlags.SIGNED_WITH_HASH]:
```
timburke: This should be `==` as well since there's a single option.

iotilecore/iotile/core/hw/reports/signed_list_format.py
```diff
@@ -23,6 +29,41 @@ class SignedListReport(IOTileReport):
     def __init__(self, rawreport, **kwargs):
         super(SignedListReport, self).__init__(rawreport, signed=True, encrypted=False, **kwargs)
```
timburke: I would recommend actually just making these static members inside the class:

```python

class SignedListReport:
    SIGNED_WITH_HASH = 0
    SIGNED_WITH_USER_KEY = 1
    SIGNED_WITH_DEVICE_KEY = 2

    _KEY_TYPE_MAP = {
        AuthProvider.NoKey: ReportSignatureFlags.SIGNED_WITH_HASH,
        ...
    }

    @classmethod
    def KeyTypeToStreamType(cls, key_type):
        stream_type = cls._KEY_TYPE_MAP.get(key_type)
        if stream_type is None:
            raise ArgumentError("Unknown Key type")
       return stream_type

# Same with the reverse direction
```

timburke: You then don't need to export another public enum and can access the values as needed on the class itself.

iotilecore/iotile/core/hw/reports/signed_list_format.py
```diff
@@ -4,12 +4,18 @@
...
+from enum import IntEnum
 from .report import IOTileReport, IOTileReading
 from iotile.core.utilities.packed import unpack
 from iotile.core.exceptions import NotFoundError, ExternalError
 from iotile.core.hw.auth.auth_provider import AuthProvider
 from iotile.core.hw.auth.auth_chain import ChainedAuthProvider
 
+class ReportSignatureFlags(IntEnum):
+    """Types of stream report signeature"""
+    SIGNED_WITH_HASH = 0
```
anwu: Typo. "signeature" should be "signature".

iotilecore/iotile/core/hw/reports/signed_list_format.py
DimaWittmann: The plan was to have a public enum, so it can be used outside of the class, for example in the tests, not sure how better to do. 
It looks like I should not have broken the compability between signature flags and key types but just document in explicitly

iotilecore/iotile/core/hw/reports/signed_list_format.py
DimaWittmann: Got it

iotilecore/iotile/core/hw/reports/signed_list_format.py
DimaWittmann: Got it

Issue #949, opened 2020-02-21
Debug Manager's flash function does not work for binary files


anwu:
When using `debug_manager.py`'s `flash` function, it states it supports flashing of binary files. However, this is not the case.

The `format_map` [here](https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/hw/debug/debug_manager.py#L260) correlates file types with certain helper functions, however the one for binary files is `None`. As a result, the check [here](https://github.com/iotile/coretools/blob/master/iotilecore/iotile/core/hw/debug/debug_manager.py#L272) raises us an `ArgumentError`.

This should be fixed so that the binary file format has a corresponding `def _process_bin`.

PR #950, opened 2020-02-21
Fix debug_manager to support flashing bin files

anwu:
## Overview & Major Changes

The `debug_manager` has a `flash` function that is meant to support flashing arbitrary firmware onto a device. This function advertises support for `.elf`, `.hex` and `.bin` files. However, currently only `.elf` and `.hex` files work.

There is a helper function associated with each file type, except for the one handling binary files. This PR is just adding the function to support `.bin` files. It is just a simple call using `IntelHex` and loading a binary file into it.

Closes https://github.com/iotile/coretools/issues/949


anwu:
> Is this imminently needed as a release, or can these changes be used straight from `master`? If there are other changes that may come along during development, consider adding to HEAD for now

Its needed for testing purposes. But I can move it into HEAD for now.

PR #957, opened 2020-03-10
Asyncio sockets in transport plugin: Fix handling of full write buffer

timothyfroehlich:
## Overview & Major Changes
On overloaded Access Points running the new plugin, I observe a long series of failed writes to the socket. This is caused by the socket buffer being full. The code should be waiting for the socket buffer to drain before attempting to write more to the buffer. No release needed at this time.


mattrunchey:
Adding @timburke. I don't know enough to say whether having the drain after the write was there for a reason. It seems fine from your testing, but best to be safe.

timothyfroehlich:
I wrote the code originally, it was like that in the examples, but it doesn't really make sense that way unless there's only a single code path / no threads / no asyncio.

PR #960, opened 2020-03-13
Add config "cloud:verify-server"

### LINKED ISSUE #952: Add cloud:verify-server config to iotile-ext-cloud
1. We need to add a new config to `iotile-ext-cloud` called `cloud:verify-server` as a bool with a default value of `True`.  
2. If set to False, this config should cause `iotile-ext-cloud` to always pass `verify=False` when creating an `Api` object to try to talk to `iotile.cloud`.

The goal is to enable self-signed certificate support in CoreTools.

### LINKED ISSUE #953: Audit codebase for any direct python-iotile-cloud `Api` usage and ensure it respects verify=False
We need to verify that we have found all of our direct use of the `Api` object in CoreTools and are not missing any places that need to updated to support `self-signed` TLS certificates.

dvksirin:
This PR adds new config option `cloud:verify-server`, value is bool, `True` by default.
This allows to disable verification of cloud server TLS certificate. It is useful when user call cloud API being behind firewall that replaces TLS certificate with self-signed one.

- added config option `cloud:verify-server`
- instantiate Api object with `verify` param
- pass `verify` param when calling cloud server with `requests`
- disable throwing `InsecureRequestWarning` when `cloud:verify-server` is `False`

Closes #952 
Closes #953

### DIFF
--- a/iotile_ext_cloud/iotile/cloud/cloud.py
+++ b/iotile_ext_cloud/iotile/cloud/cloud.py
@@ -48,10 +48,10 @@ class IOTileCloud:
         if domain is None:
             domain = self._conf.get('cloud:server')
 
-        if not self.server_cert_verifying:
+        if not self.verify_server:
             urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
-        self.api = Api(domain=domain, verify=self.server_cert_verifying, **kwargs)
+        self.api = Api(domain=domain, verify=self.verify_server, **kwargs)
         self._domain = self.api.domain
 
         try:
@@ -75,7 +75,7 @@ class IOTileCloud:
         self.token_type = self.api.token_type
 
     @property
-    def server_cert_verifying(self) -> bool:
+    def verify_server(self) -> bool:
         return self._conf.get('cloud:verify-server')
 
     def _prompt_user_pass(self, username, domain):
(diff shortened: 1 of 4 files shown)


iotile_ext_cloud/iotile/cloud/cloud.py
```diff
@@ -71,6 +74,10 @@ def __init__(self, domain=None, username=None, **kwargs):
         self.token = self.api.token
         self.token_type = self.api.token_type
```
timburke: I would recommend renaming this property to make it a little less verbose, something like `verify_server`.  Originally I was wondering why it was a property at all but I'm guessing that's to ensure that users don't try to set it on their own, which makes sense.

iotile_ext_cloud/iotile/cloud/cloud.py
```diff
@@ -43,12 +43,15 @@ class IOTileCloud:
 
     def __init__(self, domain=None, username=None, **kwargs):
         reg = ComponentRegistry()
-        conf = ConfigManager()
+        self._conf = ConfigManager()
 
         if domain is None:
-            domain = conf.get('cloud:server')
+            domain = self._conf.get('cloud:server')
```
timburke: I would recommend capturing the value of the config here and saving as a private variable `_verify`.  Since the impact of the config variable only really matters at the moment this constructor runs (since it's passed down and cached inside the created `Api` object, we shouldn't dynamically refetch the config variable when someone asks us if we're verifying the server's cert because it doesn't matter what the current value of the config object is, it only matters what the value was when the `IOTileCloud` object was created and it was cached inside `self.api`

dvksirin: Makes sense, thanks. Updated - dc76d53676

PR #973, opened 2020-04-23
Invoke python with current interpreter

anwu:
## Overview & Major Changes

The `autobuild_python_scripts` function was invoking the script using `py`. The `py` command is the Python _launcher_ that is a utility that comes with **Windows** installations. The `python` command is the actual executable that is used. On Linux, there is no `py`, unless the user modifies an alias or symlink themselves.

This was causing an issue on Codeship, since it runs its CI/CD on Linux.

So we need to the `autobuild_python_scripts` function to actually invoke the script with `python`. It should be good enough to replace `py` with `python`, but it might be better to get the actual full path to the current running Python interpreter. I believe this solution is safer, because it does not assume any environment variables to be set. Rather, it  will use `sys.executable`, which is the path of the currently running Python interpreter that is running the current script.


mattrunchey:
Ok I just didn't look far enough down the google results. `sys.executable` relies on `sys.argv[0]` https://stackoverflow.com/a/39254880/10377552 . 

I don't know enough about `iotile-build` to say whether we'd have any instance without this being present, but I imagine it's not likely.

anwu:
@mattrunchey I thought of a test that might prove/disprove whether or not this will work...
I was going to modify the script that Codeship runs as a step. Specifically, I was thinking of adding a line somewhere [in here](https://github.com/iotile/pod_controller/blob/master/.codeship-ci/setup_and_build.sh#L47) to do a `python test_sys_executable.py`. And inside the script it would try to use `subprocess` to call another script like so...
```
import sys
import os

print("attempting to run subprocess")
os.subprocess.run([sys.executable, 'subprocess_script.py'])
print("finished attempting to run subprocess")
```

And inside `subprocess_script.py` would just be something simple to indicate that it ran succesfully like
```
print("successfully called this script")
```

This way, we should be able to know with a greater level of confidence that this will work in `iotile build`.

anwu:
Cool... Did the test I mentioned above, and it worked...
```
2020-04-23 17:32:15
firmware-ci
attempting to run subprocess
2020-04-23 17:32:16
firmware-ci
successfully called script
2020-04-23 17:32:16
firmware-ci
finished attempting to run subprocess
```

PR #976, opened 2020-05-05
autobuild function to combine trub scripts

anwu:
## Overview & Major Changes

This adds a simple function in `autobuild` to combine multiple trub scripts into one.

All this does is take in two parameters, the first is a list of trub scripts that exist in the project's dependencies, and the second is the name of the output file.


x24ling:
I think for completion 
1. Add a pytest that will be run to verify that the function works
2. bump iotile-build

anwu:
> I think for completion
> 
> 1. Add a pytest that will be run to verify that the function works
> 2. bump iotile-build

Bumped the version number. Also I added a pytest that will combine two trubs, and it will check that the contents are indeed valid records.

anwu:
> test_combining_trubs passes for me. To make the test more strict, is there a way to check that each one of the records were in the ones from test_trub_1 and test_trub_2?

I made the test more strict. Most of the records have a `def __eq__` function that will let you compare the classes together. I had to update a few records though.

PR #977, opened 2020-05-07
Fix error handling in auth process

DimaWittmann:
The purpose of this PR is to fix error handling and simplify the signature of function pointers in authentication related code.

* Fixed 
* Type hints were added.
* Replaced complicated approach to provide an additional object to the functions with a particular member with a function pointer.

Test cases that particularly check this functionality:
`pytest -k test_BLE_authentication.py`

### DIFF
--- a/iotilecore/RELEASE.md
+++ b/iotilecore/RELEASE.md
@@ -3,6 +3,7 @@
 All major changes in each released version of `iotile-core` are listed here.
 
 ## HEAD
+- Fix error handling in auth process
 
 ## 5.0.15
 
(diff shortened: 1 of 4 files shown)


DimaWittmann:
> Looks good.
> 
> However, looks like I get a different issue now. I'm not sure if its related or a different issue
> 
> ```
> acceptance_tests\test_BLE_authentication.py ..Exception in thread Thread-6:
> Traceback (most recent call last):
>   File "C:\Users\AndrewWu\AppData\Local\Programs\Python\Python38-32\lib\threading.py", line 932, in _bootstrap_inner
>     self.run()
>   File "c:\arch\coretools\transport_plugins\bled112\iotile_transport_bled112\bled112_cmd.py", line 70, in run
>     callback(result_obj)
>   File "c:\arch\coretools\transport_plugins\bled112\iotile_transport_bled112\bled112.py", line 1175, in _on_authentication_finished
>     conndata['failed'] = True
> TypeError: 'NoneType' object does not support item assignment
> ```

Thank you for this information, it looks like connection was disrupted, I added required check

PR #978, opened 2020-05-08
Feat add bled broadcast mode

timothyfroehlich:
## Overview & Major Changes

This PR changes the behavior of the Bled112's Reader Thread to add an optional Broadcast V2 Deduplicator.

- The optional deduplicator, configured with the `config:bled112:deduplicate-broadcast-v2` registry setting, identifies Broadcast V2 packets using the identifying bytes in their packets and stores them. Subsequent packets are compared to the stored packet and duplicates are dropped. This happens immediately after they are read from the Bled112 dongle and before they have been repeatedly unpacked, parsed, and copied in to and read from multiple queues. This resulted in another 40% increase in performance. 
- A configurable "timeout" can be set to ensure that at least one broadcast every x seconds is allowed through.
 -I've also observed no apparent side-effects. RPCs continue to work properly and very few packets are dropped.

### Additional Notes

My tests show an extremely small amount of dropped packets, even with 75 Pods, close to the number of dropped packets that we expect due to the Bled122 switching between the channels it scans. In a five minute test, it missed no updates from a Pod that updates every second, and only one update from a Pod that updates every half second (which means that it missed three to four packets in a row from the Pod, but this only occurred once). These are massive improvements over the previous code, which missed 40 updates from the 1Hz Pod and 355 from the 0.5Hz Pod.

My tests run in real time and give me good summaries, but don't put out any easy to read graphs yet. I may enhance them to output graphs next week.


transport_plugins/bled112/iotile_transport_bled112/broadcast_v2_dedupe.py
```diff
@@ -0,0 +1,104 @@
...
+        deduper = self.dedupers.get(encoded_uuid)
+        if deduper is None:
+            deduper = BroadcastV2Deduper(encoded_uuid, self._pass_packets_every)
+            if len(self.dedupers) == self.MAX_DEDUPERS:
+                self.evict_oldest_deduper()
+            self.dedupers[encoded_uuid] = deduper
+
+        return deduper.allow_packet(data)
+
+    def evict_oldest_deduper(self):
```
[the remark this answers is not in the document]
timothyfroehlich: Thanks, I wasn't aware of the OrderedDict, I'll remember that for next time.

I'm not worried about the performance of the eviction function, It's simply a protection against memory leaks. The only situations where I could see it becoming a problem would occur when the plugin is processing incoming data from > 500 Pods at once, which I doubt we'll ever see in real life.

timothyfroehlich:
@timburke Could I get a review?

transport_plugins/bled112/iotile_transport_bled112/async_packet.py
```diff
@@ -51,37 +60,39 @@ def read_packet(self, timeout=3.0):
...
+        broadcast_v2_dedupers = BroadcastV2DeduperCollection(dedupe_timeout)
 
     while not stop.is_set():
         try:
-            header = bytearray()
-            while len(header) < header_length:
-                chunk = bytearray(filelike.read(header_length - len(header)))
-                header += chunk
+            # The bled112 will read EOF when there is no more data, so it's safe to read a large amount
+            read_buffer += bytearray(filelike.read(1024))
```
timburke, line 74: This read pattern is not a safe way to optimize this.  It trades speed in some use cases for significant slowdown and uncontrolled performance in many others.  

I suspect that you should see that if you do a test where you send RPCs or a script to a POD, you'll find the throughput is now significantly slower with this code (possibly as much as 3-4x slower) and highly variable based on how many PODs are nearby sending advertisement traffic.

This issue is that `filelike.read(1024)` blocks until either of the following conditions are met:
- there are at least 1024 bytes available OR
- the configured read timeout (set at 100ms I believe) expires and whatever bytes it has are returned

In scenarios when you are writing a command and reading the response before writing the next command, you will never have 1024 bytes filling up the buffer quickly so every command you write will now take a minimum of 100ms because you have to wait for the read to timeout before the several byte response to the command is returned to you.

In an RPC we need 2 writes (write payload, write command) and normally that takes < 1ms followed by 15-30ms for those writes to get across BLE to the POD and the response to come back.  With this change that would become 200 ms worst case for the writes since it needs to two timeouts.

The safe pattern for this kind of thing:
```python
   read_buffer += filelike.read(1)  # Is there any data at all available without assuming how much?
 
   avail_length = filelike.in_waiting  # NOTE That this is a pyserial api, not generic filelike so you may need to modify the tests

  if avail_length > 0:
      read_buffer += filelike.read(avail_length)  # This can never timeout because you just guaranteed there's enough data
```

transport_plugins/bled112/iotile_transport_bled112/async_packet.py
```diff
@@ -51,37 +60,39 @@ def read_packet(self, timeout=3.0):
...
+
+                if len(read_buffer) < next_packet_len:
+                    # Still waiting to read this packet
                     break
 
-            if stop.is_set():
-                break
+                # Process the packet and remove it from the read buffer
+                packet = read_buffer[:next_packet_len]
+                del read_buffer[:next_packet_len]
```
timburke, line 85: I don't know for sure that this is bad but I've not seen this pattern before for doing a `tail` on a buffer.  Normally I see:

```
read_buffer = read_buffer[next_packet_len:]
```

transport_plugins/bled112/iotile_transport_bled112/broadcast_v2_dedupe.py
```diff
@@ -0,0 +1,104 @@
...
+        return False
+    #This identifies the bgapi packet as an event
+    if not (packet[0] == 0x80 and packet[2] == 6 and packet[3] == 0):
+        return False
+    #This identifies the event as a broadcast v2 packet
+    if not (packet[18] == 0x1b and packet[19] == 0x16 and packet[20] == 0xdd and packet[21] == 0xfd):
+        return False
+    return True
+
+class BroadcastV2DeduperCollection():
```
timburke, line 24: Extra `()` not necessary here.

transport_plugins/bled112/iotile_transport_bled112/broadcast_v2_dedupe.py
```diff
@@ -0,0 +1,104 @@
...
+        return deduper.allow_packet(data)
+
+    def evict_oldest_deduper(self):
+        """Find and remove the oldest deduper
+
+        This function will likely be called rarely, if ever
+        """
+        oldest = list(self.dedupers.values())[0]
+
+        for deduper in iter(self.dedupers.values()):
```
timburke, line 70: `iter` is not needed here:

```
for deduper in self.dedupers.values():
```

timburke: Actually given that the goal of this function is just to ensure the dict doesn't grow without bound, I would recommend a simpler eviction policy:

```python
def __init__(self):
    self.dedupers = OrderedDict()

def evict_oldest_deduper(self):
    first = next(iter(self.dedupers))  # From https://stackoverflow.com/a/21067850/9739119
  
    # Always delete the old pod that was added rather than finding out which pod is the least recently seen
    # if the pod is still active, it will come back at the end of the list and eventually the top of the dict will
    # be those pods that are seen rarely.
    del self.dedupers[first]
```

transport_plugins/bled112/iotile_transport_bled112/config_variables.py
```diff
@@ -9,5 +9,8 @@ def get_variables():
     conf_vars.append(["throttle-broadcast", "bool", "Only report changing broadcast values, not all values", "false"])
     conf_vars.append(["throttle-scan", "bool", "Only report device_seen once in a timeout", "false"])
     conf_vars.append(["throttle-timeout", "int", "device_seen and report events timeout in seconds", 30])
+    conf_vars.append(["deduplicate-broadcast-v2", "bool", "Filter unchanging Broadcast V2 packets per device", "false"])
```
timburke, line 12: Since this will be a public api we can't change, I think we should use a cleaner naming scheme:

```
deduplicate-advertisements   # We're filtering all advertisements not just those containing broadcast data
deduplicate-timeout
```

transport_plugins/bled112/iotile_transport_bled112/async_packet.py
```diff
@@ -51,37 +60,39 @@ def read_packet(self, timeout=3.0):
...
 
-            remaining_length = length_function(header)
+                if broadcast_v2_dedupers:
```
timburke, line 87: This can be expressed more simply as:
```python
if broadcsat_v2_dedupers is not None and broadcast_v2_dedupers.allow_packet(packet):
    continue
```

python gurantees left-to-right evaluation of conditions so this is safe and `broadcast_v2_dedupers` will not be accessed if its None.

transport_plugins/bled112/iotile_transport_bled112/async_packet.py
```diff
@@ -24,7 +26,14 @@ def __init__(self, filelike, header_length, length_function):
         self.file = filelike
         self._stop = Event()
 
-        self._thread = Thread(target=ReaderThread, args=(filelike, self.queue, header_length, length_function, self._stop))
+        dedupe = False
```
timburke, line 29: To keep the interface to this class clean, I would recommend moving these assignments out into the main `BLED112DeviceAdapter` class and just passing in kwargs of `dedupe` and `dedupe_timeout` with defaults:

```
def __init__(..., deduplicate=False, deduplicate_timeout=0)

transport_plugins/bled112/iotile_transport_bled112/async_packet.py
timothyfroehlich: Sounds good. I was a bit surprised when I didn't notice any performance issues, and I most likely didn't perform the right test to see them. `in_waiting` looks great, I'll implement that.

transport_plugins/bled112/iotile_transport_bled112/async_packet.py
timothyfroehlich: I didn't see any apparent issues from it, but I can change it.

transport_plugins/bled112/iotile_transport_bled112/broadcast_v2_dedupe.py
timothyfroehlich: Ok, sure.

transport_plugins/bled112/iotile_transport_bled112/config_variables.py
timothyfroehlich: Well, we're not deduplicating all BLE advertisements, just the advertisements that include the bytes that identify them as an IOTile broadcast_v2 advertisement. I'm certainly open to a better name though.

transport_plugins/bled112/iotile_transport_bled112/async_packet.py
timothyfroehlich: Confirmed, pulling a large amount of text from a device, using an RPC sequence that requires one RPC request and response per 20 bytes of text, took about 20% longer with the current code. (though not 3-4x slower as predicted). Tests where both the sender and reciver used the code from this PR took 30% longer. Will test with the changes @timburke suggested.

transport_plugins/bled112/iotile_transport_bled112/async_packet.py
timothyfroehlich: True. Isn't `is not None` the same as `if broadcast_v2_dedupers` though?

timburke: > Isn't `is not None` the same as `if broadcast_v2_dedupers` though?

More or less.  Technically the first check just if it's assigned explicitly to `None` whereas the second allows any value that evaluates to True, which can lead to some surprises but in this case they're equivalent.  

I personally like the explicitness of comparing to `None` when that's the intent but people differ on that one.

The comment was mainly just about combining the two `if` statements together.

transport_plugins/bled112/iotile_transport_bled112/config_variables.py
timburke: Maybe:
`throttle-v2-advertisements`
`throttle-v2-timeout`

One quick comment on `v1`, vs `v2`.  The version applies to the **advertisement packet format**, whether or not that advertisement contains a broadcast value or not.  So you can have a v2 advertisement without broadcasting anything and it would still be throttled by your code as I understand it.

timburke:
@timothyfroehlich One more thing to check is that this still works when `active-scan` is set to True.  In that case, I believe, even though the v2 advertisements don't have scan response information, the bled112 sends a packet with all `0s` in it as a scan response since the host computer can't know whether to expect a scan response or not.  

There are two potential issues:
1. Those scan responses get through and cut into the performance improvement because there are still a lot of them coming through
2. The scan-response parsing logic in the BLED112 adapter code expects that a scan-response will match a previously seen advertisement.  I'm not sure if it behaves properly if one slips through whereas the advertisement was filtered out.

timothyfroehlich:
> @timothyfroehlich One more thing to check is that this still works when `active-scan` is set to True. In that case, I believe, even though the v2 advertisements don't have scan response information, the bled112 sends a packet with all `0s` in it as a scan response since the host computer can't know whether to expect a scan response or not.

> 
> There are two potential issues:
> 
> 1. Those scan responses get through and cut into the performance improvement because there are still a lot of them coming through

- The AP is still running with active scan and I've seen no issues with it, my performance numbers were captured while all APs were running with active scan.

> 2. The scan-response parsing logic in the BLED112 adapter code expects that a scan-response will match a previously seen advertisement.  I'm not sure if it behaves properly if one slips through whereas the advertisement was filtered out.

- It appears that it's fine, it drops a scan response that it wasn't waiting for here: 
https://github.com/iotile/coretools/blob/c3f9e3dd599c121fbe63878ef5726fcf9c843b96/transport_plugins/bled112/iotile_transport_bled112/bled112.py#L785

This does all point to potential performance increases by disabling active scan, but I believe that was tried before. Perhaps not with my current tester.

transport_plugins/bled112/iotile_transport_bled112/async_packet.py
timothyfroehlich: Sure, I've already adjusted the if statement. Thanks for the clarification, having the more explicit check is definitely better.

transport_plugins/bled112/iotile_transport_bled112/config_variables.py
timothyfroehlich: Ok, that sounds good.

transport_plugins/bled112/iotile_transport_bled112/async_packet.py
timothyfroehlich: The adjustment looks good, I'm seeing similar times to read 5500 characters from an AP on all versions again. 

I'm working on upgrading my performance tests so I'll have to wait to see how this affects advertisement performance.

transport_plugins/bled112/iotile_transport_bled112/broadcast_v2_dedupe.py
timothyfroehlich: Got a test failure from that, looks like insertion order wasn't guaranteed until Python 3.6. Using the `OrderedDict` that Dima suggested instead.

mattrunchey: Mostly unimportant side info: technically we aren't *supposed* to rely on inherent dictionary ordering until Python 3.7 . It's considered an "implementation detail" in 3.6, so other versions of Python aren't guaranteed to support it (i.e. `pypy` or `Jython` might not compile that same requirement in to their interpreter/language as `Cython`, the reference implementation).

Interestingly, some places claim insertion order to be maintained all the way back to 3.5, but I more commonly see it cited in 3.6 .

http://gandenberger.org/2018/03/10/ordered-dicts-vs-ordereddict/ exemplifies the behavior difference in `IronPython` and even `pandas`. 

There is more gory discussion about this on the python-dev mailing list, https://mail.python.org/pipermail/python-dev/2016-September/146327.html . It was officially made a core feature rather than an implementation detail later, https://mail.python.org/pipermail/python-dev/2017-December/151283.html

Anyway, this comment is mostly just fun reading if anyone was curious about what is happening under the hood here :)

timothyfroehlich:
I'm seeing some not-good behavior related to the changes I've made to the serial read code. When the device is overwhelmed with incoming data, and the deduplication of broadcasts is disabled, the device no longer processes any packets, they all end up in buffers that increase until the program crashes. Ending the simulation allows the device to begin processing those buffered packets and forwarding them. I'm going to work on my simulator and test harness so I can more easily run tests, and make sure that these read loop changes are worth it.

timothyfroehlich:
@timburke Now that I've finished up my AP simulation suite, I'm going to use it to begin testing this change, including long-term runs.

For my tests, I'm going to run the APs against various numbers of Pods, checking for a few things:
Number of packets dropped from a specific Pod, the presence of any packet delivery lag, and whether every update from a Pod that updates every 0.6 seconds makes it through. These numbers will be compared to AP's running AP 0.9.3 and 0.10.3, representing Coretools from last year and Coretools as of a few months ago. I'll set the tests to run for a few hours if possible.

I'm also going to run a "worst case scenario" test where every Pod updates it's value with every broadcast and see if performance has dropped in this (unrealistic) scenario.

And, I'll run a second check to make sure that the RPC performance is still good, by timing how long it takes to read 5000+ characters from an AP.

Can you think of more tests that you'd like to see?

timothyfroehlich:
I've reverted the read loop changes. In my performance tests I did see some slight performance gains from it but overall I judged it not good enough to make the change. I'm looking to merge this in now.

PR #979, opened 2020-05-12
Add CRC calculation of sensor graph to config variable

anwu:
## Overview & Major Changes

The main goal of this PR is to provide the user/programmer a way to calculate the CRC of an entire sensor graph (nodes, streamers, constants, configs, metadata). The purpose of calculating a CRC is so that it can provide us a unique identifier of that particular sensor graph, despite whatever the app tag says the version is.

This is my approach to the usage and how I implemented it...
In a `.sgf` file, the user needs to set a config variable to the controller with the type of algorithm/polynomial used for the CRC checksum. The file will also need a `meta` variable called  `set_crc` with an rvalue of the address of the config variable that was specified.

Here's an example
```
meta set_crc = 0x9999;

config controller
{
    set 0x9999 to 0x104C11DB7 as uint32_t;
}
```
The `.sgf` file essentially tells the sensor graph compiler to calculate a CRC with the polynomial, `0x104C11DB7`, and to update the config variable at `0x9999` with that calculated CRC. 

Here are the main changes made...

**SensorGraph**
- New class method called `add_crc` that can be called regardless if the `.sgf` specified a CRC variable
- Some `dump` commands to help build a binary representation of the instance
- Method called `calculate_sensorgraph_crc` to calculate a CRC over the binary representation of the instance

**Parser**
- Added a line to call `SensorGraph.add_crc()`. This line will do nothing if the user's `.sgf` doesn't specify CRC config info

### Additional Notes

The sensor graph's compiler is a very important and sensitive piece of technology. It was very important that I maintained as much firmware agnosticism as possible. Also, it was very important for me to make this as backwards and forwards compatible as possible.

There was a lot of thought that went into this. More information on the discussion and reasoning behind the design could be found in the respective Twist post.


iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -317,6 +317,21 @@ def get_tick(self, name):
...
+    def add_crc(self):
+        """Check metadata to see if the sensorgraph's CRC should be added."""
+
+        if 'set_crc' not in self.metadata_database:
+            return
+
+        slot = SlotIdentifier.FromString('controller')
+        config_id = self.metadata_database['set_crc']
+        config_type, polynomial = self.get_config(slot, config_id)
+        calculated_crc = self.calculate_sensorgraph_crc(polynomial=polynomial)
```
ekkizogloy, line 329: I would change calculate_sensorgraph_crc to a more generic call that takes an identifier defining the algorithm to use like "CRC32_104C11DB7", "SHA256" or "CRC16". Internally the generic function would use the appropriate algorithm as needed.

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -514,3 +529,41 @@ def dump_streamers(self):
...
+    def calculate_sensorgraph_crc(self, polynomial=0x104C11DB7):
+        """Returns the calculated CRC of the sensor graph"""
+
+        sensor_graph_dump = self.dump_nodes() + self.dump_streamers() +\
+                            self.dump_constant_database() +\
+                            self.dump_metadata_database() +\
+                            self.dump_config_database()
+
+        sensor_graph_string = ', '.join(sensor_graph_dump)
+        sensor_graph_bytes = bytes(sensor_graph_string, 'utf-8')
```
ekkizogloy, line 557: While this is the way to calculate the checksum of a file, the bytes representation of the 'utf-8' characters will not be available in a POD. There must be another representation that can be used for this such that a POD can be verified without the SGF file.

iotilesensorgraph/iotile/sg/parser/parser_v1.py
```diff
@@ -112,6 +112,7 @@ def compile(self, model):
             statement.execute(self.sensor_graph, self._scope_stack)
 
         self.sensor_graph.initialize_remaining_constants()
+        self.sensor_graph.add_crc()
         self.sensor_graph.sort_nodes()
```
ekkizogloy, line 116: I'm not sure, however we would likely want the configuration variable nodes to be added after all the nodes have been sorted. (Unsure exactly what this would do to the representation)

iotilesensorgraph/iotile/sg/graph.py
anwu: Sure. I was thinking of changing it to something like `def calculate_hash(self, algorithm)`. And I guess we'll have to build some sort of known list of algorithms so that this function can match and use the appropriate one.

So its function would do something like
```
def calculate_hash(self, algorithm):
    hash_algorithm = KNOWN_ALGORITHMS[algorithm]
    hash_algorithm.calculate(binary_representation)
```

Maybe have a class like
```
class CRC:
    def __init__(self, polynomial, crc_size):
        self.polynomial = polynomial
        self.crc_size = crc_size
    def calculate(self, binary_representation):
        # actual algorithm
```

ekkizogloy: Yea that's what I'm thinking... You could make a class that provides all the different algorithms however I will leave that to you if that's what you want to do. 

Along with the known list of algorithms, there would be an integer identifier that gets set in the configuration variable. My guess is that for the first implementation, we would have a single algorithm (this CRC32 that you have coded) and it's representation can be 0x01.

iotilesensorgraph/iotile/sg/parser/parser_v1.py
anwu: Yeah it rearranges how the nodes are in its list. So that would affect it.

I will fix this to `add_crc` after sorting.

iotilesensorgraph/iotile/sg/graph.py
anwu: I think the objective/goal is getting a bit convoluted here. Sounds like we are on a different page.

I thought that the intention with the config variable was to hold the calculated checksum of the `.sgf` file, and that the variable should be basically read only when a programmer wants to see that information so that they can compare it with an `.sgf` they have.

Are you saying with your statement that the firmware is now responsible for dumping out all of its sensor graph's content, so that the sensor graph python proxy can reconstruct the `SensorGraph` instance? or at least spit out a binary representation?

Lets try this exercise... I want to try to understand the use case that you are expecting from this feature. How do you see yourself or other programmers verify a device's sensor graph contents? What are the commands you would execute?
`iotile hw --port bled112 connect <uuid> controller sensor_graph calculate_checksum`
Where `calculate_checksum` returns a checksum after sending RPC's to the device to dump its sensor graph? And from there the user can compare that calculated checksum with the actual `.sgf`?

What I had in mind was just looking at the config variable and just trusting it for what it is.
`iotile hw --port bled112 connect <uuid> controller config_manager get_variable 0x9999`
And the user would just compare that value with the actual `.sgf`, or whatever list of checksums we have that we've associated with certain sensorgraphs.

anwu: I think we need to first talk about what we want a binary representation of.

If its just the sensorgraph nodes, then [it already exists](https://code.eng.archsys.io/github.com/iotile/pod_controller@6d77dfa6bac718ea73e4934d57d9ebbb8f51091d/-/blob/lib_controller/python/support/sensorgraph.py#L134:9).
If we want it to include constants, firmware will need to know how to dump `rsl` constants.
And if we want config variables, we will need it to call RPC's to the config manager, which is a different interface. And do we want to include metadata in the binary representation as well?

anwu:
### Changes (Updated)
- Updated where hashing algorithm is called
- Flexible hashing algorithms
- Added helper functions to get binary representations of sensorgraph elements
- Fixed checksum calculation to not include the checksum variable in itself

## How To Test
Take any `.sgf` file and add both meta tags to specify the hash checksum address, as well as the hash algorithm.
```
meta app_tag = 2048;
meta app_version = "1.2";
meta hash_address = 0x2100;
meta hash_algorithm = "sha256";

... the rest of the sgf ...
```
This will tell the sensor graph compiler to use the `sha256` algorithm to calculate the checksum over the various sensor graph elements, and to store that string in the config variable at `0x2100`.

In the `.sgf` file, you will need to specify a line that sets a config variable on the controller to indicate that it is the config variable to store the checksum. This address must match the address specified in the meta tag of `meta hash_address`.

```
config controller
{
    set 0x2100 to "none" as string;
}
```
When this entry's config variable matches the address in `meta hash_address`, it tells the sensor graph compiler that the checksum will be saved to `0x2100` as a string.

Right now, all the hashing algorithm will return a string.

### Example Test
1. Update any production application's `.sgf` file with appropriate meta tags and config variable entries. You can try the following...
```
meta hash_address = 0x2100;
meta hash_algorithm = "sha256";

config controller
{
    set 0x2100 to "none" as string;
}
```
2. Compile the sensor graph with verbose settings.
`iotile-sgcompile <path_to_sgf> -f snippet -v`
3. Verify that you see debug statements with the different checksums like so...
```
DEBUG:iotile.sg.graph:nodes_checksum: be9aad6e365394784e5bd7ba709a7763552740441cd047a33d68891aed1f8f66
DEBUG:iotile.sg.graph:streamers_checksum: ce72710023cfcd9d70114caee3066438851b82ec4ccbe09c5eddd13abf07e46e
DEBUG:iotile.sg.graph:configs_checksum: dbfde31c13513b96d9b7e93e198746d0f638c6aa590837f0934569cbd6bb7548
DEBUG:iotile.sg.graph:constants_checksum: 9845adde3fae1d9378bef6ca1fd28bffb659ec9c5d2c1fb8a667c65e24ccbf0a
DEBUG:iotile.sg.graph:metadata_checksum: 61292e044eefbbe7c7827b3287360ce8d7424fd1662e68025c27913f22b4342d
DEBUG:iotile.sg.graph:device_checksum: 54fbbca1724e10e29c12be2e2611419ca9b5ec0bf6bd25358093b20a2ebef7f6
```

iotilecore/iotile/core/utilities/hash_algorithms/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+from .sha import calculate_sha
+
+HashAlgorithm = namedtuple("HashAlgorithm", ['calculate', 'algorithm'])
+
+KNOWN_HASH_ALGORITHMS = {
+    'crc32_0x104C11DB7': HashAlgorithm(calculate_crc, 0x104C11DB7),
+    'sha256': HashAlgorithm(calculate_sha, 256)
+}
+
+__all__ = ['KNOWN_HASH_ALGORITHMS']
```
anwu, line 14: This `__init__.py` file is where we want to list frequently used and known hashing algorithms. This package will allow a programmer to easily select the proper hashing functions based on the index they provide to the dict.

For example, they can select the sha256 algorithm by doing this
```
hash_alg = KNOWN_HASH_ALGORITHMS['sha256']
hash_alg.calculate(hash_alg.algorithm, some_bytes)
```

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -317,6 +318,45 @@ def get_tick(self, name):
         except ArgumentError:
             return 0
 
+    def add_checksum(self):
+        """Check metadata if sensorgraph's checksum needs to be added."""
+
+        if 'hash_address' not in self.metadata_database or\
+            'hash_algorithm' not in self.metadata_database:
+            return
```
anwu, line 326: User must provide BOTH meta tags `hash_address` as well as `hash_algorithm` in order to add a checksum to the `.sgf`.

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -317,6 +318,45 @@ def get_tick(self, name):
...
+        self._logger.debug("configs_checksum: %s", configs_checksum)
+        constants_checksum = hash_algorithm.calculate(hash_algorithm.algorithm,
+                                                  self.get_constant_database_binary())
+        self._logger.debug("constants_checksum: %s", constants_checksum)
+        metadata_checksum = hash_algorithm.calculate(hash_algorithm.algorithm,
+                                                  self.get_metadata_database_binary())
+        self._logger.debug("metadata_checksum: %s", metadata_checksum)
+        combined_checksum_string = nodes_checksum + streamers_checksum +\
+                                   configs_checksum + constants_checksum +\
+                                   metadata_checksum
```
anwu, line 351: Note that the order here is important in calculating the entire device's checksum. This must be the same order in which the proxy to the device calculates its checksum.

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -317,6 +318,45 @@ def get_tick(self, name):
...
+        combined_checksum_bytes = bytes(combined_checksum_string, 'utf-8')
+        device_checksum = hash_algorithm.calculate(hash_algorithm.algorithm, 
+                                                   combined_checksum_bytes)
+        self._logger.debug("device_checksum: %s", device_checksum)
+
+        config_type, _ = self.get_config(slot, config_id)
+        self.add_config(slot, config_id, config_type, device_checksum)
```
anwu, line 358: This replaces the value of the `config_id` variable, which the user specifies as the address that matches the `meta hash_address` value.

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -514,3 +559,102 @@ def dump_streamers(self):
...
+    def dump_config_database(self, dump_config_type=True, ignore_configs=[]):
+        """Dump all of the config variables in this sensor graph as a list of strings."""
+
+        config_dump = []
+        for slot, conf_vars in self.config_database.items():
+            for conf_var, conf_def in conf_vars.items():
+                conf_type, conf_val = conf_def
+
+                if conf_var in ignore_configs:
+                    continue
```
anwu, line 592: We typically want to ignore the actual config variable that holds our checksum calculation when we are doing the checksum calculation. If we don't ignore this config variable, we will have a chicken and egg problem when trying to figure out the checksum.

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -514,3 +559,102 @@ def dump_streamers(self):
...
+
+                if dump_config_type:
+                    config_dump.append("'{}' {} {} {}".format(slot, conf_var,
+                                                              conf_type,
+                                                              conf_val))
+                else:
+                    config_dump.append("'{}' {} {}".format(slot, conf_var,
+                                                              conf_val))
```
anwu, line 600: We want to ignore the actual data type of the variable, since we have no way of knowing what exact type is going to be returned on the proxy side. The only information we can get back from the proxy side about the variable is the length of the data type.

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -514,3 +559,102 @@ def dump_streamers(self):
...
+    def get_metadata_database_binary(self):
+        """Returns the binary representation of the KNOWN metadata variables"""
+
+        binary_representation = bytearray()
+
+        for metadata in self.dump_metadata_database():
+            entry, _ = metadata.split(': ')
+
+            if entry in known_metadata:
+                binary_representation += bytes(metadata, 'utf-8')
```
anwu, line 646: At this moment, we only want to support the two known meta tags, `app_tag` and `app_version`.

iotilebuild/iotile/build/config/site_scons/trub_script.py
```diff
@@ -367,6 +368,30 @@ def combine_trub_scripts(trub_scripts_list, out_file):
...
+
+        for record in trub_script.records:
+            if isinstance(record, SetConfigRecord) and\
+            record.config_id == checksum_address:
+                device_checksum = record.data
+                return device_checksum
+
+    return None
+
+
```
anwu, line 394: Searches through a list of TRUB scripts and looks for a record where the config id matches the checksum address.

anwu:
## More Changes
- Fixed [this issue](https://github.com/iotile/coretools/issues/980) where the `iotile-sgcompile -f scripts` output would not work properly for config variables that were arrays
- Added meta tag parsing for `hash_configs_ignorelist` and `hash_configs_ignorelist_address`. The compiler will parse the configs at the address and exclude them from the checksum calculation.

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -317,6 +319,101 @@ def get_tick(self, name):
...
+            'hash_algorithm' not in self.metadata_database or\
+            'hash_algorithm_address' not in self.metadata_database or\
+            'hash_configs_ignorelist' not in self.metadata_database or\
+            'hash_configs_ignorelist_address' not in self.metadata_database:
+            return
+
+        slot = SlotIdentifier.FromString('controller')
+        hash_address = self.metadata_database['hash_address']
+        algorithm = self.metadata_database['hash_algorithm']
+        algorithm_address = self.metadata_database['hash_algorithm_address']
```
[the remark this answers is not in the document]
anwu: That is what I had done initially. However, @ekkizogloy pointed out that the implementation is not too obvious, and it can be limiting. I see validity in both perspectives, and I feel like its just preference at this point.

iotilesensorgraph/iotile/sg/graph.py
```diff
@@ -317,6 +319,101 @@ def get_tick(self, name):
         except ArgumentError:
             return 0
 
+    def _parse_configs_ignorelist(self):
```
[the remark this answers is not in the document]
anwu: That's a good idea. I can do this.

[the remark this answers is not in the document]
anwu: Fixed

iotilesensorgraph/RELEASE.md
```diff
@@ -5,6 +5,7 @@ here.
 
 ## HEAD
```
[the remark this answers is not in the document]
anwu: Fixed

anwu:
Changed the format of the `meta` tag for `ignorelist`.

Before
```
meta hash_configs_ignorelist = "controller:0xcafe,controller:0xbabe,slot 1:0x5555";
```
Now
```
meta hash_configs_ignorelist = "'controller':[0x2100,0x21AD,0x21AE],'slot 1':[0x5555]"
```
Basically now it is a stringified dictionary. The programmer can choose to encapsulate the data with '{}' or not.

PR #987, opened 2020-07-14
Add tqdm to the utc assigner

ekkizogloy:
Add the tqdm statusbar to the utc_assigner. It is imported using a `try {} except {}` block and not included as a dependency in the setup.py. 

This is needed because there are some conditions where the utc assignment takes a significant amount of time. This status bar is needed to know that software is operating correctly and not stalled.

### DIFF
--- a/iotilecore/iotile/core/hw/reports/utc_assigner.py
+++ b/iotilecore/iotile/core/hw/reports/utc_assigner.py
@@ -27,15 +27,6 @@ from typedargs.exceptions import ArgumentError, KeyValueException
 from .signed_list_format import SignedListReport
 from .report import IOTileReading
 
-try:
-    from tqdm import tqdm
-except ImportError:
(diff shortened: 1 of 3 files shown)


ekkizogloy:
@timburke Is there anyway that we can fast track this for KLA?

iotilecore/iotile/core/hw/reports/utc_assigner.py
```diff
@@ -23,7 +23,7 @@
 import datetime
 import logging
 from bisect import bisect_left
-from typedargs.exceptions import ArgumentError
+from typedargs.exceptions import ArgumentError, KeyValueException
```
timburke, line 26: extra `KeyValueException` import is no longer needed.

ekkizogloy: removed

PR #994, opened 2020-09-28
Add multiple stream support to bled112 deduplication

timothyfroehlich:
## Overview & Major Changes

Enhances the early broadcast v2 deduplication in the bled112 transport to be aware of Pods that broadcast multiple streams. Each stream will be deduplicated separately. Without this change and if the Pod rotated the stream it broadcasts quickly enough, essentially no deduplication would take place, leading to performance losses.


timburke (approved):

timothyfroehlich:
@timburke there's something wrong with the api hook to travis. It looks like it's passing. Do you think you could merge this for me?
https://travis-ci.org/github/iotile/coretools/builds/732040263

timburke:
> It looks like it's passing. Do you think you could merge this for me?

Done

PR #995, opened 2020-09-30
Feat add inmemory authenticator

anwu:
## Overview & Major Changes

We need a way to support applications interfacing `coretools` to be able to authenticate their devices without being hung on a CLI prompt.
Currently, if a user wants to connect to a device, they must either set an environment variable `USER_KEY_<device MAC> = <key>` OR if it doesn't exist, be prompted by CLI to enter a password manually. This implementation is an issue for any application that needs to interface the password protection feature without getting blocked.

The solution was to create a new authentication provider that would allow the user to call the `InMemoryAuthProvider`'s class method to set shared passwords that the authentication logic can retrieve. See my comments in the review for a walkthrough of all the changes.

- Added a mapping of devices' connection strings and UUID to the adapter
- Added a new authentication provider
- Fixed authentication to support authenticating via UUID or MAC
- Added `iotile registry` support to add a temporary password
```
(venv_py38) PS C:\Arch\Repositories> iotile registry
(AnnotatedRegistry) set_temporary_password 0xa51 test
(AnnotatedRegistry) back
(root) hw --port bled112
(HardwareManager) connect 0xa51
```

### DIFF
--- a/iotilecore/iotile/core/dev/annotated_registry.py
+++ b/iotilecore/iotile/core/dev/annotated_registry.py
@@ -83,7 +83,7 @@ class AnnotatedRegistry(object):
 
     @param("device_uuid", "int", desc="Device UUID")
     @param("password", "str", desc="Password to the device")
-    def set_temporary_password(self, device_uuid, password: str):
+    def set_temporary_password(self, device_uuid: int, password: str):
         InMemoryAuthProvider.add_password(device_uuid, password)
 
     @param("device_uuid", "int", desc="Device UUID")
--- a/iotilecore/iotile/core/dev/annotated_registry.py
+++ b/iotilecore/iotile/core/dev/annotated_registry.py
@@ -3,6 +3,7 @@
 # tool.  Since the registry is used internally in the type system it cannot
 # itself make use of typedargs annotations
 from iotile.core.utilities.typedargs import annotated, param, return_type, context
+from iotile.core.utilities.typedargs.annotate import docannotate
 from iotile.core.exceptions import ExternalError
 from iotile.core.hw.auth.inmemory_auth_provider import InMemoryAuthProvider
 from .registry import ComponentRegistry
@@ -81,13 +82,23 @@ class AnnotatedRegistry(object):
 
         return self.reg.list_config()
 
-    @param("device_uuid", "int", desc="Device UUID")
-    @param("password", "str", desc="Password to the device")
+    @docannotate
     def set_temporary_password(self, device_uuid: int, password: str):
+        """Sets a temporary password for a device in memory for the current instance.
+
+        Args:
+            device_uuid (int): the device UUID
+            password (str): the password to save in memory
+        """
         InMemoryAuthProvider.add_password(device_uuid, password)
 
-    @param("device_uuid", "int", desc="Device UUID")
+    @docannotate
     def clear_temporary_password(self, device_uuid: int):
+        """Clears the password based on the device UUID
+
+        Args:
+            device_uuid (int): the device UUID
+        """
         InMemoryAuthProvider.clear_password(device_uuid)
 
     @param("key", "string")
--- a/iotilecore/iotile/core/dev/annotated_registry.py
+++ b/iotilecore/iotile/core/dev/annotated_registry.py
@@ -11,14 +11,18 @@ from .registry import ComponentRegistry
 _name_ = "Developer"
 
 
-@annotated
+@docannotate
 def registry():
+    """Returns an Annotated Registry
+
+    Returns:
+        AnnotatedRegistry show-as context: an Annotated Registry
+    """
     return AnnotatedRegistry()
 
 @context()
 class AnnotatedRegistry(object):
-    """
-    AnnotatedRegistry
+    """AnnotatedRegistry
 
     A mapping of all of the installed components on this system that can
     be used as build dependencies and where they are located.  Also used
@@ -28,21 +32,25 @@ class AnnotatedRegistry(object):
     def __init__(self):
         self.reg = ComponentRegistry()
 
-    @param("component", "path", "exists", desc="folder containing component to register")
+    @docannotate
     def add_component(self, component):
-        """
-        Register a component with ComponentRegistry.
+        """Register a component with ComponentRegistry.
 
         Component must be a buildable object with a module_settings.json file that
         describes its name and the domain that it is part of.
+
+        Args:
+            component (path): folder containing component to register
         """
 
         self.reg.add_component(component)
 
-    @return_type("map(string, string)")
+    @docannotate
     def list_plugins(self):
-        """
-        List all of the plugins that have been registerd for the iotile program on this computer
+        """List all of the plugins that have been registerd for the iotile program on this computer
+        
+        Returns:
+            map(string, string): List of plugins registered for this iotile instance
         """
 
         return self.reg.list_plugins()
@@ -50,34 +58,39 @@ class AnnotatedRegistry(object):
     def find_component(self, key, domain=""):
         return self.reg.find_component(key, domain)
 
-    @param("key", "string", desc="iotile component to find")
+    @docannotate
     def remove_component(self, key):
-        """
-        Remove component from registry
+        """Remove component from registry
+
+        Args:
+            key (str): iotile component to remove
         """
 
         return self.reg.remove_component(key)
 
-    @annotated
+    @docannotate
     def clear_components(self):
-        """
-        Clear all of the registered components
+        """Clear all of the registered components
         """
 
         self.reg.clear_components()
 
-    @return_type("list(string)")
+    @docannotate
     def list_components(self):
-        """
-        List all of the registered component names
+        """List all of the registered component names
+
+        Returns:
+            list(string): list of the registered component names
         """
 
         return self.reg.list_components()
 
-    @return_type("list(string)")
+    @docannotate
     def list_config(self):
-        """
-        List all of the registered component names
+        """List all of the registered config names
+
+        Returns:
+            list(string): list of registered config names
         """
 
         return self.reg.list_config()
@@ -101,22 +114,38 @@ class AnnotatedRegistry(object):
         """
         InMemoryAuthProvider.clear_password(device_uuid)
 
-    @param("key", "string")
-    @param("value", "string")
+    @docannotate
     def set(self, key, value):
+        """Sets a config value in the registry to a value
+
+        Args:
+            key (str): the config variable
+            value (str): the value to be set
+        """
         self.reg.set_config(key, value)
 
-    @param("key", "string")
-    @return_type("string")
+    @docannotate
     def get(self, key):
+        """Gets a config value from its key in the registry
+
+        Args:
+            key (str): the config variable
+
+        Returns:
+            string: the value of the config variable
+        """
         return self.reg.get_config(key)
 
-    @param("key", "string")
-    @return_type("string")
+    @docannotate
     def clear(self, key):
+        """Clears a config key from the registry
+
+        Args:
+            key (str): the config variable
+        """
         self.reg.clear_config(key)
 
-    @annotated
+    @docannotate
     def freeze(self):
         """Freeze the current list of extensions to a single file.
 
@@ -130,7 +159,7 @@ class AnnotatedRegistry(object):
 
         self.reg.freeze_extensions()
 
-    @annotated
+    @docannotate
     def unfreeze(self):
         """Remove any frozen extension list."""
 
--- a/iotilecore/iotile/core/dev/annotated_registry.py
+++ b/iotilecore/iotile/core/dev/annotated_registry.py
@@ -4,6 +4,7 @@
 # itself make use of typedargs annotations
 from iotile.core.utilities.typedargs import annotated, param, return_type, context
 from iotile.core.exceptions import ExternalError
+from iotile.core.hw.auth.inmemory_auth_provider import InMemoryAuthProvider
 from .registry import ComponentRegistry
 
 _name_ = "Developer"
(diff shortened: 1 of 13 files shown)


iotilecore/iotile/core/dev/annotated_registry.py
```diff
@@ -80,6 +81,15 @@ def list_config(self):
...
 
+    @param("device_uuid", "int", desc="Device UUID")
+    @param("password", "str", desc="Password to the device")
+    def set_temporary_password(self, device_uuid, password: str):
+        InMemoryAuthProvider.add_password(device_uuid, password)
+
+    @param("device_uuid", "int", desc="Device UUID")
+    def clear_temporary_password(self, device_uuid: int):
+        InMemoryAuthProvider.clear_password(device_uuid)
+
```
anwu, line 92: Allows user to set a temporary password to connect to a device within the same hardware instance.

iotilecore/iotile/core/hw/auth/auth_chain.py
```diff
@@ -44,12 +45,23 @@ def _load_installed_providers(self):
         for name, entry in reg.load_extensions('iotile.auth_provider'):
             self._auth_factories[name] = entry
 
+    def _get_device_alias(self, device_id):
```
anwu, line 48: Looks for either a connection string or uuid based on the argument.

iotilecore/iotile/core/hw/auth/auth_chain.py
```diff
@@ -58,7 +70,11 @@ def get_root_key(self, key_type, device_id):
             try:
                 return provider.get_root_key(key_type, device_id)
             except NotFoundError:
-                pass
+                try:
```
anwu, line 73: Attempt to get the root key with the same provider if an alias for the device exists. (If using UUID failed, use MAC)

iotilecore/iotile/core/hw/auth/inmemory_auth_provider.py
```diff
@@ -0,0 +1,63 @@
+from iotile.core.exceptions import NotFoundError
+from .rootkey_auth_provider import RootKeyAuthProvider
+from .auth_provider import AuthProvider
+
+
+class InMemoryAuthProvider(RootKeyAuthProvider):
+    """Key provider implementation that allows user to set a password in memory"""
+
+    _shared_passwords = {}
```
anwu, line 9: @timburke Wondering if we should limit the size of `_shared_passwords`...

transport_plugins/bled112/iotile_transport_bled112/bled112_cmd.py
```diff
@@ -299,11 +326,9 @@ def notified_payload(event):
 
         supported_auth = AuthType.AUTH_METHOD_0.value | AuthType.AUTH_METHOD_1.value \
                          | AuthType.AUTH_METHOD_2.value | AuthType.AUTH_METHOD_3.value
-        permisions = 0x00
-        token_gen = 0x01
 
-        manager = BLED112AuthManager(permisions, token_gen)
-        success, data = manager.authenticate(device_uuid, supported_auth, send_auth_client_request)
+        self._auth_manager._key_provider.conn_map = self._conn_map
```
anwu, line 330: Gives the connection map over to the authentication chain so that it can look for device aliases if needed.

anwu:
[CI passed](https://travis-ci.org/github/iotile/coretools), the hook is failing to report properly.
Looks like Travis CI is [having some issues with GitHub API requests](https://www.traviscistatus.com/).

transport_plugins/bled112/iotile_transport_bled112/bled112_cmd.py
timburke: In terms of what part of the system does what, I would recommend refactoring this:
- the auth system is designed to work via device ids, not connection strings / mac addresses so I would not leak the connection map over to that system.
- instead, you can just do the lookup inside the bled112 code between `connection_string` and `device_id` and then only ever hand a `device_id` over to the auth system so it doesn't need to worry about it.
- if you can't find the right device id in your conn_map then it fails the authentication procedure without asking `_auth_manager` for anything since it doesn't have the right prerequisites (no device id).

transport_plugins/bled112/iotile_transport_bled112/bled112_cmd.py
```diff
@@ -26,6 +29,8 @@ def __init__(self, stream, commands, stop_check_interval=0.01):
         self._logger = logging.getLogger(__name__)
         self._logger.addHandler(logging.NullHandler())
         self.event_handler = None
+        self._conn_map = []
+        self._auth_manager = BLED112AuthManager(0x00, 0x01)
```
timburke, line 33: I'm not sure its safe to save this off here.  `BLED112CommandProcessor` is an multi-ble-connection system.  What happens if we are authenticating to 2 PODS at once?  

Any state that pertains to a single connection should move to the `BLED112DeviceAdapter` code where we have a dict for keeping things on a per-connection basis.  Look, for example, at how `IOTileReportParser` is created and handled, which is also a per-connection state machine.  

Similarly, I would recommend moving `_conn_map` to `BLED112DeviceAdapter` as well.  There is already a place in the code where `BLED112DeviceAdapter` sets a `uuid` for the connection that is ultimately passed through to the AuthChain.  By having `_conn_map` in that class instead, you can just pass the correct `device_id` at that stage vs passing a connection_string to `BLED112CommandProcessor` and then having it try and convert it to a device_id before using it.

transport_plugins/bled112/iotile_transport_bled112/bled112_cmd.py
```diff
@@ -78,6 +83,30 @@ def run(self):
                 self._logger.exception("Fatal error in background processing loop")
                 raise
 
+    def update_conn_map(self, conn_string, device_uuid):
```
timburke, line 86: Same comment as above, this behavior can stay inside of `BLED112DeviceAdapter` and it should simplify the code.

timburke (commented):
See some potential race issues and ways to minimize state leakage across different classes that don't all need to know about this change.

transport_plugins/bled112/iotile_transport_bled112/bled112_cmd.py
anwu: Thanks for the suggestion. Implemented in this latest push.

iotilecore/iotile/core/hw/auth/inmemory_auth_provider.py
timburke: No, it can't fill up on its own so we shouldn't limit the user in how many entries they choose to put there.

iotilecore/iotile/core/dev/annotated_registry.py
```diff
@@ -80,6 +81,15 @@ def list_config(self):
 
         return self.reg.list_config()
 
+    @param("device_uuid", "int", desc="Device UUID")
+    @param("password", "str", desc="Password to the device")
+    def set_temporary_password(self, device_uuid: int, password: str):
+        InMemoryAuthProvider.add_password(device_uuid, password)
```
timburke, line 87: There should be docstrings for this since that sets the `help` text when running `help set_temporary_password` in the iotile shell.

Also, all new annotated functions should use `@docannotate` instead of `@param`.  That removes the need to duplicate `desc=` strings and improves how the help text looks.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -113,6 +114,7 @@ def __init__(self, port, on_scan=None, on_disconnect=None, passive=None, **kwarg
         self.count_lock = threading.Lock()
         self.connecting_count = 0
         self.maximum_connections = 0
+        self._conn_map = []
```
timburke, line 117: There's a better data structure choice here than a list.

You want an `OrderedDict`.  It comes with `popitem(last=False)` to remove the oldest entry.  When setting a new entry, you can move it to end using `move_to_end` if it already exists:

```python
_conn_map = OrderedDict()

def _update_conn_map(conn_string, device_uuid):
    while len(_conn_map) >= ConnMapMaxSize:
        _conn_map.popitem(last=False)

    if conn_string in _conn_map:
        _conn_map.move_to_end(conn_string)
    else:
        _conn_map[conn_string] = device_uuid
```

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -184,6 +186,38 @@ def reset_scan_stats(self):
...
+        entry = (conn_string, device_uuid)
+
+        if entry in self._conn_map:
+            self._conn_map.remove(entry)
+        elif len(self._conn_map) >= self.ConnMapMaxSize:
+            self._conn_map = self._conn_map[-(self.ConnMapMaxSize-1):]
+
+        self._conn_map.append(entry)
+
+    def _get_device_alias(self, device_id):
```
timburke, line 213: I don't think you should ever need a bidirectional lookup to implement this feature.  If you do, then there's a bug somewhere else in the existing implementation that should be fixed instead of working around it with a bidirectional lookup.

Can you verify:
1. `context['uuid']` is only ever used in the authentication related functions.  If not, lets discuss where else it is used.  The below assumes it was added previously to enable this feature.

Given the above, the underlying problem is that `context['uuid']` is incorrectly set in the code.  It should have always been the `device_id` not the mac address/connection string.  The fix is to do the `connection_string -> device_id` lookup before setting `context['uuid']` and just set it to `device_id` if known and `None` if the lookup fails.

During authentication then, `context['uuid']` is always either the correct device_id or a `None`.  If auth is required and `context['uuid'] is None` then auth fails, otherwise it proceeds.

There should never need to be a lookup of connection string from `device_id`.  Let me know if I'm missing something in the code.  Everywhere inside of bled112 device adapter knows the connection string already, it's just the device id that the code doesn't know.

timburke (changes requested):
Overall plan looks good but there are some implementation questions still.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
anwu: Gotcha... yeah I was looking at `OrderedDict` but i didn't see a `move_to_end` function. I must have been looking at an older version.

timburke: The other standard option is just to `pop` and `add` so it ends up at the end anyway.

iotilecore/iotile/core/dev/annotated_registry.py
anwu: Got it. Fixed it. Can I fix the rest of this file? Or are there reasons we would want to have those remain as is?

timburke: No reason they need to remain as is, if it's easy to update.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
anwu: Got it. I thought it was a bit odd and wondered if we should just drop using the MAC as a potential look up. I looked through and tested the code and verified the `context['uuid'] is only used in those authentication functions. I went ahead and fixed where that `connection_string` was initially fed into the chain of callbacks (inside of `_probe_characteristics_finished`).
I also am just using the standard `get` function on the dictionary. That will automatically return `None` if it doesn't exist, which will get caught inside the check that I added in the `BLED112AuthManager`.

iotilecore/iotile/core/dev/annotated_registry.py
anwu: Cool, updated it.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
```diff
@@ -184,6 +187,29 @@ def reset_scan_stats(self):
...
+        Args:
+            conn_string (str): The string of the device's MAC address
+            device_uuid (int): The integer representation of the device's UUID
+        """
+        if conn_string in self._conn_map:
+            self._conn_map.move_to_end(conn_string, last=True)
+        else:
+            self._conn_map[conn_string] = device_uuid
+
+        while len(self._conn_map) >= self.ConnMapMaxSize:
```
timburke, line 210: Not super important but the behavior with this order of operations is not super intuitive.  Consider if `ConnMapMaxSize == 1`.  It would result in never filling the conn_map even though the user specified a size of 1.

If you put the `while` at the head of the function you free up space and then add a new entry.  Alternatively the conditional should be `>` not `>=` if you have it in the current location.

transport_plugins/bled112/iotile_transport_bled112/bled112_auth.py
```diff
@@ -164,12 +164,15 @@ def authenticate(self, uuid, client_supported_auth, handshake_fn: SendAuthHandsh
...
 
         Args:
-            uuid (int or str): either device mac or id
+            uuid (int): the device UUID
             auth_type (int): See .AuthType
             handshake_fn (SendAuthHandshakeFn): a function pointer
         Returns:
             (bool, bytearray/dict): tuple of success flag, session key or dict with reason of failure
         """
+        if not uuid or not isinstance(uuid, int):
```
timburke, line 173: this should be `uuid is None` instead of `not uuid`.   `not uuid` returns True if `uuid == 0`, which is a potentially valid device_id you shouldn't reject.

timburke (approved):
Looks good to me, see the two minor cleanup things.

transport_plugins/bled112/iotile_transport_bled112/bled112.py
anwu: I'll just change it to `>` at the current location. 
The reason I chose to put that at the end was because I wanted to make sure that we don't pop the oldest value in the scenario that we aren't adding a new entry, but just moving an entry to the end.
The alternative to fix that logic with having it at the head of the function would require us to do a `while len(self._conn_map) > self.ConnMapMaxSize and conn_string not in self._conn_map`.

transport_plugins/bled112/iotile_transport_bled112/bled112_auth.py
anwu: Good catch. Fixed

PR #996, opened 2020-10-27
Fix memory leak in the Operation Manager

timothyfroehlich:
## Overview & Major Changes

An incorrect comparison to None prevented the list of waiter callbacks from being cleaned up when the callback was ready to call back. The result was a monotonically increasing memory leak, seen in the AP which was using the Operation Manager inside the Supervisor.

### DIFF
--- a/iotilecore/iotile/core/utilities/async_tools/operation_manager.py
+++ b/iotilecore/iotile/core/utilities/async_tools/operation_manager.py
@@ -145,7 +145,7 @@ class OperationManager:
         del loc[OperationManager._LEAF]
 
         for parent, key in reversed(parents):
-            if len(parent[key]) is not None:
+            if len(parent[key]) > 0:
                 return
 
             del parent[key]


timburke (approved):

timothyfroehlich:
@timburke is it normal for Travis to be waiting this long to launch the build? It's still in the queue.

mattrunchey:
@timothyfroehlich if you search on the travis CI community forums, you will see that it is a more widespread issue https://travis-ci.community/t/build-delays-for-open-source-project/10272/3 . 

On that forum they suggest that this is due to Travis's shutdown of travis-ci.org . They recommend we move all of our open source builds to travis-ci.com (posts in that thread suggest that will resolve this issue). https://travis-ci.community/t/build-delays-for-open-source-project/10272/13

timothyfroehlich:
@mattrunchey What do you suggest I do then? I don't feel that I should be making changes to coretools's testing infrastructure, I barely know it.

mattrunchey:
I would encourage you to not be afraid to upgrade testing infrastructure -- you won't break anything :) 

You might be able to look at a previous PR where we moved the OSX and Windows tests to Azure. For linux, it may be as trivial as copying the Azure pipelines commands and replacing the OS it runs on without any other changes. https://github.com/iotile/coretools/pull/704

PR #1001, opened 2021-05-25
Add linux to tested platforms

dmaone:
as a part of migration to the common CI platform: github actions

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,9 +5,9 @@ on:
     branches: [ master ]
 
 jobs:
-  test-windows:
-    name: test-windows py3${{ matrix.python-minor }}
-    runs-on: windows-2019
+  test-linux:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -5,9 +5,9 @@ on:
     branches: [ master ]
 
 jobs:
-  test-windows:
-    name: test-windows py3${{ matrix.python-minor }}
-    runs-on: windows-2019
+  test-linux:
+    name: test linux py3${{ matrix.python-minor }}
+    runs-on: ubuntu-16.04
```
mattrunchey, line 10: Didn't you want to use at least ubuntu 18?

mattrunchey (approved):

.github/workflows/test.yml
dmaone: I can't - locally-stashed qemu doesn't work, and I couldn't find the replacement quickly.

PR #1005, opened 2021-07-01
Move deployment to github actions.

dmaone:
Finally we can retire travis and consolidate all our CI on github actions.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Validate package
+        id: validate
+        run: python scripts/validate_and_get_relnotes.py >> $GITHUB_ENV
+        env:
+          EVENT: ${{ toJSON(github.event) }}
+
+      - uses: pypa/gh-action-pypi-publish@v1.4.2
+        with:
+          password: ${{ secrets.PYPI_API_TOKEN }}
+          # Uncomment these for debugging and publishing to test PyPI
```
mattrunchey, line 49: Are there other lines that should be commented here? Looks like it's still pointing at test.pypi.org.

mattrunchey (approved):

.github/workflows/deploy.yml
dmaone: Spot-on, haha! All 3 should be commented

